PR #15, opened 2021-07-17
Tweaks for SubModule usage and Mac Debugging

baconpaul:
1. Add some interface targets to CMakeLists.txt so
   when clap is used as a submoudle you can add interface
   linkage to your target. Loads more to do here of course

2. Add a better error mesage in one spot in the host

3. On MacOS QT the QTApplication doesn't keep command line
   arguments properly. We should resovle this of course, but for
   now, add an environment variable which overrides the arguments
   to allow me to at least debug my plugin in clion on mac.


abique (approved):

abique:
Thank you :)

PR #44, opened 2022-01-03
Fix struct definition for clap_plugin_factory / clap_plugin_factory_t

AlexHarker:
This commit fixes the definition to a typedef and defines clap_plugin_factory_t, matching the style of the rest of the SDK.

### DIFF
--- a/include/clap/plugin-factory.h
+++ b/include/clap/plugin-factory.h
@@ -13,7 +13,7 @@ extern "C" {
 
 // Every methods must be thread-safe.
 // It is very important to be able to scan the plugin as quickly as possible.
-struct clap_plugin_factory {
+typedef struct clap_plugin_factory {
    /* Get the number of plugins available.
     * [thread-safe] */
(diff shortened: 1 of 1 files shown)


abique:
This was already fixed in the branch next.

AlexHarker:
Ah - sorry - I didn't see that - I thought next had only a few changes. Should I be working against next?

AlexHarker:
Is this in your local copy only? I see nothing up here?

abique:
It is in https://github.com/free-audio/clap/blob/next/include/clap/entry.h
I think I'll merge next into main in one or two days.

AlexHarker:
This fix isn't for clap_entry it is for clap_plugin_factory which doesn't seem correct to me.

abique:
Ah indeed!
Then yes your pull request is totally valid!
Thanks and sorry for being confused.

AlexHarker:
No worries.

PR #63, opened 2022-03-03
Add avendish to projects which support clap

jcelerier:
It's not *entirely* finished yet but most of it is there. Avendish is a C++ media plug-in library which has the particularity of being non-intrusive, it works entirely through reflection: plug-ins don't even need to know about the binding library (although helper types are provided). This gives much more freedom to the plug-in author than in usual plug-in apis: it's possible to write plug-ins that work per-sample, per mono buffers, per arrays of channels, etc. with a fairly large freedom of expression. It's also possible to make non-audio things, for instance 2D texture, Max/Pd-like message processors, etc.

Here are a couple examples in the lib: 

- A most basic audio plugin (as you can see, there's exactly *zero* dependency, no inheritance, no nada :-)):
```C++
#pragma once
#include <cmath>
struct TanhDisto {
  static constexpr auto name() { return "Tanh Distortion"; }
  static constexpr auto c_name() { return "tanh_distortion"; }
  static constexpr auto uuid() { return "63d5b569-e1e5-4f2d-9f84-0bd710689b19"; }
  static constexpr int input_channels() { return 2; }
  static constexpr int output_channels() { return 2; }

  void operator()(double** inputs, double** outputs, int frames)
  {
    using namespace std;
    for(int c = 0; c < input_channels(); ++c) {
      for(int k = 0; k < frames; k++) {
        outputs[c][k] = std::tanh(inputs[c][k]);
      }
    }
  }
};
```

- A plug-in with a control and some helper types: 

https://github.com/celtera/avendish/blob/main/examples/Helpers/Lowpass.hpp

- A plug-in which processes things per-sample: 

https://github.com/celtera/avendish/blob/main/examples/Modular.hpp

- A minimal synth:

https://github.com/celtera/avendish/blob/main/examples/Midi.hpp

cheers !

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ and use to get a basic plugin experience:
 
 - [clap-juce-extension](https://github.com/free-audio/clap-juce-extension), juce add-on
 - [MIP2](https://github.com/skei/MIP2), host and plugins
+- [Avendish](https://github.com/celtera/avendish), a reflection-based API for media plug-ins in C++ which supports Clap
 
 ## Programming Language Bingings
 


baconpaul:
What a neat project! (I don’t have merge here but I just scanned avendish and it is very clever)

abique:
Thanks :)

Issue #70, opened 2022-04-01
Meaning of CLAP_AUDIO_PORTS_PREFERS_64BITS


glowcoil:
The meaning of the `CLAP_AUDIO_PORTS_PREFERS_64BITS` flag seemed a bit vague to me and I wanted to clarify.

If the `CLAP_AUDIO_PORTS_PREFERS_64BITS` flag is true for an audio port, does that mean the host must provide a 64-bit buffer, or that the plugin will support either 32- or 64-bit and simply prefers 64-bit? Likewise, if the `CLAP_AUDIO_PORTS_PREFERS_64BITS` flag is false for an audio port, does that mean the host must provide a 32-bit buffer, or that the plugin will support either and simply prefers 32-bit?

The comment at the top of `audio-ports.h` specifies that if a plugin does not implement the audio ports extension, it only supports 32-bit audio, which makes sense. But if a plugin does implement the extension, can it still specify that it only supports 32-bit audio by setting `CLAP_AUDIO_PORTS_PREFERS_64BITS` to false on all its ports? Or does implementing the audio ports extension mean that the plugin must unavoidably support both 32- and 64-bit audio, and can only specify a "preference"?

robbert-vdh:
I asked the same thing in https://github.com/free-audio/clap/pull/71#discussion_r841730054, so I'll repeat the comment here for visilibity:

> In which situation would this preference [referring to `CLAP_AUDIO_PORTS_PREFERS_64BITS`] be useful? Wouldn't hosts either use 32-bit audio internally and pass that audio to plugins as is, or use 64-bit audio in their audio graph and either convert it to and from 32-bit for plugins that don't support 64-bit audio or pass that to plugins as is? The host converting its own 32-bit audio to 64-bit for a plugin wouldn't be very useful, in that case the plugin could just do the conversion themselves (and save on a buffer memcpy).

abique:
It is required to support 32 bits, always.
supports means that 64 can be used.
preferred means that ideally the host should use a 64 bit audio buffer but this is not a requirement.

baconpaul:
There isn't a supports_64 flag though is there?

So all ports must support 32 bit fine
but they can also support or support-and-prefer 64
the former means daw option, use float if you can
the second means daw option, use double if you can
and there is no version for must-use-double because not all daws support it?

but if that's the case aren't we missing a flag?

glowcoil:
[...]

PR #71, opened 2022-04-04
clap 0.24.0

abique:


include/clap/ext/audio-ports.h
```diff
@@ -26,8 +27,11 @@ enum {
    // Main port must be at index 0.
    CLAP_AUDIO_PORT_IS_MAIN = 1 << 0,
 
-   // The prefers 64 bits audio with this port.
-   CLAP_AUDIO_PORTS_PREFERS_64BITS = 1 << 1,
+   // The port can be used with 64 bits audio
+   CLAP_AUDIO_PORT_SUPPORTS_64BITS = 1 << 1,
+
+   // 64 bits audio is preferred with this port
+   CLAP_AUDIO_PORTS_PREFERS_64BITS = 1 << 2,
```
robbert-vdh, line 34: In which situation would this preference be useful? Wouldn't hosts either use 32-bit audio internally and pass that audio to plugins as is, or use 64-bit audio in their audio graph and either convert it to and from 32-bit for plugins that don't support 64-bit audio or pass that to plugins as is? The host converting its own 32-bit audio to 64-bit for a plugin wouldn't be very useful, in that case the plugin could just do the conversion themselves (and save on a buffer memcpy).

include/clap/ext/params.h
```diff
@@ -84,40 +84,21 @@ enum {
...
-   // Does this param supports per note automations?
-   CLAP_PARAM_IS_PER_NOTE = 1 << 1,
-
-   // Does this param supports per channel automations?
-   CLAP_PARAM_IS_PER_CHANNEL = 1 << 2,
-
-   // Does this param supports per port automations?
-   CLAP_PARAM_IS_PER_PORT = 1 << 3,
-
    // Useful for for periodic parameters like a phase
```
robbert-vdh, line 87: `for for`. Also, might be worth documenting how this works. Is this the same as VST3's wraparound thing? (where values outside of `[0, 1]` should be treated as ``x `fmod` 1.0``? This is a documentation thing though, so it's probably better to go through all of this once before 1.0 since there a couple other things that could be documented better.

robbert-vdh (commented):
Really like the `gui.h` changes. The VST3-like `request_resize()` -> `set_size()` if unsuccessful is unfortunate but luckily these resizes would seldomly be denied so in practice this probably won't cause any real problems.

include/clap/ext/audio-ports.h
abique: Maybe the host would instead produce a 64 bit audio in certain situation instead.
Also imagine that you put the plugin in a modular system, you can then have the info that the port prefers 64 bit audio and then connect a 64 bit stream to it. We have situations like this in Bitwig's device builder.

include/clap/ext/params.h
abique: Yes, it is like fmod: `((value - min) % (max - min)) + min`

include/clap/ext/audio-ports.h
robbert-vdh: Wouldn't that be incredibly inefficient? You either have a 64-bit stream, or you don't. If the audio data was originally 32-bit (since there's not really any reason to use 64-bit audio on the audio graph level), then the plugin would be much better off doing the conversion to 64-bit itself if it really needs 64-bit audio internally.

robbert-vdh: #70 also asks about this, so I'll repeat the question there. Please ignore this thread!

abique: In bitwig device builder, there are components which requires 64 bit audio on certain ports because they process the input in way were the precision matters.
When we're patching cables, we get the visual info that this port wants 64 bit audio, and this hint helps us to make some decision.
So the info is also useful at the human level.

PR #75, opened 2022-04-25
Allow CLAP_EXPORT to have a custom definition

jpcima:
If building a plugin which is statically wrapped into another format, the symbol `clap_entry` is not desired to be visible.

This patch allows to prevent it by passing `-DCLAP_EXPORT=`.

### DIFF
--- a/include/clap/private/macros.h
+++ b/include/clap/private/macros.h
@@ -1,17 +1,19 @@
 #pragma once
 
 // Define CLAP_EXPORT
-#if defined _WIN32 || defined __CYGWIN__
-#   ifdef __GNUC__
-#      define CLAP_EXPORT __attribute__((dllexport))
+#if !defined(CLAP_EXPORT)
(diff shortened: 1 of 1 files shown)


abique:
Thank you again!
Sorry we're not setup yet regarding CLA.
We sort this out and I'll merge your patch.
Regards,
Alex

robbert-vdh:
Since CLAP is MIT licensed you wouldn't need a CLA, right? Or are there any plans to relicense it? I'd say having mixed copyright would actually cement its status as an open API, since it wouldn't be possible for the owner to some day decide to relicense everything under an incompatible license without getting approval from the contributors.

abique:
I want to guarantee that all contributions are made as MIT.
Maybe something like that: https://elinux.org/Developer_Certificate_Of_Origin
We are looking into it.

jpcima:
@abique Since it's a trivial modification, can you possibly author it yourself?
I am not interested in the authorship.

PR #91, opened 2022-06-06
Update README.md

Trinitou:
Would be glad to be part of this amazing project ... even if it's a small commit ;)


abique:
Thank you! ;-)

PR #99, opened 2022-06-15
CMakeLists: Handle older compilers; target trimming

baconpaul:
1. The 'compile_cpp' for c 17 doesn't work on MSVC 2019 or older gccs
2. These tests and the like should only be included if building clap
directly

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,14 +10,9 @@ target_include_directories(clap-core INTERFACE include)
 
 install(DIRECTORY include DESTINATION "." OPTIONAL EXCLUDE_FROM_ALL)
 
-get_directory_property(parent_dir PARENT_DIRECTORY)
-if ("${parent_dir}" STREQUAL "")
-    set(CLAP_IS_TOPLEVEL 1)
-else ()
-    set(CLAP_IS_TOPLEVEL 0)
-endif ()
+option(CLAP_BUILD_TESTS "Should CLap build tests and the like?" OFF)
 
-if (${CLAP_IS_TOPLEVEL})
+if (${CLAP_BUILD_TESTS})
     message(STATUS "Including clap tests, compile tests, and versions")
     add_custom_target(clap-tests)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 3.17)
 enable_testing()
 project(CLAP C CXX)
 
-add_custom_target(clap-tests)
-
 # If you use clap as a submodule of your plugin you need some interface projects
 # to allow you to link. clap-core gives you the include directory and clap-plugin-core
 # gives you the core + plugin-glue.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -12,53 +10,66 @@ target_include_directories(clap-core INTERFACE include)
 
 install(DIRECTORY include DESTINATION "." OPTIONAL EXCLUDE_FROM_ALL)
 
-add_executable(clap-compile-c EXCLUDE_FROM_ALL src/main.c)
+get_directory_property(parent_dir PARENT_DIRECTORY)
+if ("${parent_dir}" STREQUAL "")
+    set(CLAP_IS_TOPLEVEL 1)
+else ()
+    set(CLAP_IS_TOPLEVEL 0)
+endif ()
```
robbert-vdh, line 18: I'll strongly advice against doing it this way. I've seen other projects do similar things (the VST3 SDK is also notorious for it), and it causes a ton of issues and makes things difficult to work with. Instead, I'd just gate this behind a define (e.g. `CLAP_COMPILE_TESTS`), and then pass that option to the GitHub Actions CI with `-DCLAP_COMPILE_TESTS=1`. That's a lot more resilient and doesn't rely on filesystem locations and CMake's internals.

robbert-vdh (changes requested):

CMakeLists.txt
baconpaul: Great happy to make that change.

CMakeLists.txt
```diff
@@ -10,14 +10,9 @@ target_include_directories(clap-core INTERFACE include)
...
-get_directory_property(parent_dir PARENT_DIRECTORY)
-if ("${parent_dir}" STREQUAL "")
-    set(CLAP_IS_TOPLEVEL 1)
-else ()
-    set(CLAP_IS_TOPLEVEL 0)
-endif ()
+option(CLAP_BUILD_TESTS "Should CLap build tests and the like?" OFF)
 
-if (${CLAP_IS_TOPLEVEL})
+if (${CLAP_BUILD_TESTS})
```
robbert-vdh, line 15: The `-DCLAP_BUILD_TESTS=1` will also need to somehow be added to https://github.com/free-audio/clap/blob/main/.github/workflows/cmake.yml#L39 (not familiar with this action, I'd just invoke CMake manually myself).

robbert-vdh (changes requested):

CMakeLists.txt
```diff
@@ -12,53 +10,61 @@ target_include_directories(clap-core INTERFACE include)
...
-    if (${CMAKE_C_COMPILER_ID} STREQUAL "GNU" OR ${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
-        target_compile_options(clap-compile-${SUFFIX} PRIVATE -Wall -Wextra -pedantic)
-    endif()
+    add_executable(clap-compile-c EXCLUDE_FROM_ALL src/main.c)
 
-    if (${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
-        target_compile_options(clap-compile-${SUFFIX} PRIVATE -Werror=pragma-pack)
-    endif()
-endmacro()
+    macro(clap_compile_cpp SUFFIX EXT STDC STDCPP)
```
abique, line 21: Could we detect if a C or C++ standard is supported here?

CMakeLists.txt
```diff
@@ -12,53 +10,61 @@ target_include_directories(clap-core INTERFACE include)
...
-    target_link_libraries(clap-plugin-template PRIVATE -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/src/linux-my_plug.version)
-    target_link_libraries(clap-plugin-template PRIVATE -Wl,-z,defs)
-    set_target_properties(clap-plugin-template PROPERTIES SUFFIX ".clap" PREFIX "")
-elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    target_link_options(clap-plugin-template PRIVATE -exported_symbols_list ${CMAKE_CURRENT_SOURCE_DIR}/src/macos-symbols.txt)
+    clap_compile_cpp(c11    c 11 11)
+    clap_compile_cpp(cpp11 cc 11 11)
+    clap_compile_cpp(cpp14 cc 11 14)
+    # This configuration isn't supported by MSVC or older gccs so leave it off for now
+    # clap_compile_cpp(c17    c 17 17)
```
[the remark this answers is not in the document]
baconpaul: That's correct. I should look how to get better detection. I just primarily wanted us back in the water!

CMakeLists.txt
```diff
@@ -12,53 +10,61 @@ target_include_directories(clap-core INTERFACE include)
 
 install(DIRECTORY include DESTINATION "." OPTIONAL EXCLUDE_FROM_ALL)
 
-add_executable(clap-compile-c EXCLUDE_FROM_ALL src/main.c)
+option(CLAP_BUILD_TESTS "Should CLap build tests and the like?" OFF)
```
abique, line 13: Typo here: `CLap` instead of `CLAP`

CMakeLists.txt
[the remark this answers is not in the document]
abique: We can skip the detection if we rely upon `CLAP_BUILD_TESTS`.

CMakeLists.txt
abique: This action is using `CMakePresets.json`, so you could pass the option there.

CMakeLists.txt
abique: Let's not do detection it is painful and not worth it.
All we need is the github action to cover the tests.

abique:
If we make 1.0.2, then we also need to edit the changelog and version.h.

PR #108, opened 2022-06-17
Update plugin-template.c

plgDavid:
features need NULL Termination


abique:
Absolutely!
Thank you very much :-)

PR #110, opened 2022-06-18
Added MultitrackStudio and Qtractor to readme

HammyHavoc:
Figured it's best to keep this up to date for when devs check the repo.


baconpaul:
Thanks for submitting this

We think we are going to maintain the list of things which support CLAP somewhere outside the clap repo and point to it from the README so we don't have to rev the clap version for every new plugin which supports it. We are working through this as of late Friday so we *probably* won't merge this (also qtractor isn't quite done yet but good eyes from my KVR post :) )

So my guess is we won't merge this but will merge an equivalent repoint.

Thanks!

abique:
Let's merge this one. And yes we're not going to continue to list the hosts and plugins in the readme because it is already growing so quickly!

Thank you @HammyHavoc

abique:
PS: I've rebased your changes onto the branch next.

PR #113, opened 2022-06-20
entry.h: Clarify recursion stopping points

baconpaul:
Clarify the .clap file type on OS as bundle vs directory
and subsequent host stopping behavior per discussion in
https://github.com/free-audio/clap/discussions/102

### DIFF
--- a/include/clap/entry.h
+++ b/include/clap/entry.h
@@ -29,9 +29,12 @@ extern "C" {
 // of this writing).
 //
 // Each directory should be recursively searched for files and/or bundles as appropriate in your OS
-// ending with the extension `.clap`. On windows and linux, the '.clap' file is a renamed shared
-// library object, so traversing the path should stop at regular files or symlinks. On macOS
-// the '.clap' file is a bundle, so traversing the path should stop at directories.
+// ending with the extension `.clap`.
+//
+// On Windows and Linux, the '.clap' file is a renamed shared library object, so traversing the path
+// should stop at the first non-directory (regular file, symlink, etc...) whose name ends in '.clap'.
+// On macOS the '.clap' file is a bundle, so traversing the path should stop at directories whose
+// name ends in '.clap'.
 //
 // Every method must be thread-safe.
 typedef struct clap_plugin_entry {
--- a/include/clap/entry.h
+++ b/include/clap/entry.h
@@ -29,9 +29,11 @@ extern "C" {
 // of this writing).
 //
 // Each directory should be recursively searched for files and/or bundles as appropriate in your OS
-// ending with the extension `.clap`.
+// ending with the extension `.clap`. On windows and linux, the '.clap' file is a renamed shared
+// library object, so traversing the path should stop at regular files or symlinks. On macOS
+// the '.clap' file is a bundle, so traversing the path should stop at directories.
 //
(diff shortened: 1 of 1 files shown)


include/clap/entry.h
```diff
@@ -29,9 +29,11 @@ extern "C" {
 // of this writing).
 //
 // Each directory should be recursively searched for files and/or bundles as appropriate in your OS
-// ending with the extension `.clap`.
+// ending with the extension `.clap`. On windows and linux, the '.clap' file is a renamed shared
+// library object, so traversing the path should stop at regular files or symlinks. On macOS
```
abique, line 33: Why this sentences? `so traversing the path should stop at regular files or symlinks`

abique: You mean that it should not follow symlinks if they are pointing to directories?

include/clap/entry.h
```diff
@@ -29,9 +29,11 @@ extern "C" {
 // of this writing).
 //
 // Each directory should be recursively searched for files and/or bundles as appropriate in your OS
-// ending with the extension `.clap`.
+// ending with the extension `.clap`. On windows and linux, the '.clap' file is a renamed shared
```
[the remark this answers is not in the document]
baconpaul: pushed an update resolving both comments. Thank you.

abique:
I'm not sure if the text regarding traversing directories actually helps.
It makes the documentation harder to understand, while I'm not sure if it brings much to the table.

I believe that the rule is already clear:
- scan recursively
- `*.clap` is the plugin file/bundle

abique:
Did you stumble upon a problem that required to further specify this behavior?

include/clap/entry.h
robbert-vdh: In that case it should be "stop at regular files or symlinks to regular files ending in `.clap`".

abique: Stop what?

abique: Sorry but I've read that paragraph many times and I'm super confused by it, also I don't understand what problem it tries to solve.

robbert-vdh:
> Did you stumble upon a problem that required to further specify this behavior?

In #102 there was some confusion about whether or not `~/.clap/foo.clap/foo/bar/foo.clap` or `~/.clap/foo.clap/Contents/x86_64-linux/foo.clap` was a legal directory structure. I don't see why there would be any reason to believe it isn't or any any host could try to dlopen() that entire `foo.clap` directory (since that simply won't happen in any directory walking implementation), but that's what triggered this.

baconpaul:
> Did you stumble upon a problem that required to further specify this behavior?

yes

we accidentally installed surge so that it installed in "Surge XT.clap/Surge XT.clap" and bitwig didn't scan it.

robbert-vdh:
> we accidentally installed surge so that it installed in "Surge XT.clap/Surge XT.clap" and bitwig didn't scan it in windows

Has that been reported on the interop tracker yet?

baconpaul:
> > we accidentally installed surge so that it installed in "Surge XT.clap/Surge XT.clap" and bitwig didn't scan it in windows
> 
> Has that been reported on the interop tracker yet?

we fixed surge so i didn't think to report it on interoperability. i will.

PR #131, opened 2022-07-06
Threading clarifications

tim-janik:
This branch records some clarifications on CLAP's threading ideas.
It's based on recent discord discussions and comments by @abique, thanks for that.

### DIFF
--- a/include/clap/ext/thread-check.h
+++ b/include/clap/ext/thread-check.h
@@ -18,8 +18,9 @@ extern "C" {
  * to meet realtime requirements. I.e. only carry out sufficiently performant and time-bound
  * operations. So mutexes, memory (de-)allocations, IO, UI rendering should generally be avoided.
  *
- * Depending on the host, [main-thread] and [audio-thread] may or may not map to the same
- * system thread. So while plugins should use the thread checking functions and may assert
+ * Depending on the host scheduler, the [audio-thread] may move from one OS thread to another,
+ * including the same OS thread as the [main-thread].
+ * So while plugins are encouraged to use the thread checking functions and may assert
  * is_main_thread() == true or is_audio_thread() == true in a particular context, it should
  * be avoided to assert that the current thread is *NOT* is_main_thread() or is_audio_thread().
  */
--- a/include/clap/ext/thread-check.h
+++ b/include/clap/ext/thread-check.h
@@ -11,7 +11,8 @@ extern "C" {
 /* A note on threads as understood by CLAP:
  *
  * In the [main-thread], a CLAP plugin may carry out allocations, acquire a mutex and do IO,
- * bascially anything a reasonably performing program could do.
+ * basically anything a reasonably performing program could do. Long running tasks such as
+ * indexing presets, should still be run in background threads to keep the [main-thread] responsive.
  *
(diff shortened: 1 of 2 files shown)


include/clap/ext/thread-check.h
```diff
@@ -8,9 +8,24 @@ static CLAP_CONSTEXPR const char CLAP_EXT_THREAD_CHECK[] = "clap.thread-check";
 extern "C" {
 #endif
 
+/* A note on threads as understood by CLAP:
+ *
+ * In the [main-thread], a CLAP plugin may carry out allocations, acquire a mutex and do IO,
```
abique, line 13: Add something like long running tasks such as indexing presets, ... should be run in background threads to keep the main thread responsive.

include/clap/ext/thread-check.h
```diff
@@ -8,9 +8,24 @@ static CLAP_CONSTEXPR const char CLAP_EXT_THREAD_CHECK[] = "clap.thread-check";
...
+ * bascially anything a reasonably performing program could do.
+ *
+ * Within an [audio-thread] (of which there may be many in a given host), plugins should strive
+ * to meet realtime requirements. I.e. only carry out sufficiently performant and time-bound
+ * operations. So mutexes, memory (de-)allocations, IO, UI rendering should generally be avoided.
+ *
+ * Depending on the host, [main-thread] and [audio-thread] may or may not map to the same
```
abique, line 20: I would change it to:

* There can be only one main-thread and audio-thread at a time
* The audio-thread may move from one OS thread to another, including the same OS thread as the main-thread

abique (commented):
In general I'm in favor of concise documentation, where all the implication can be guessed with a bit of thinking. Like how they documented the single unix specification and posix.

I think that over explaining something leads to:
- why is the same thing explained twice? Is there an exception?
- the text is too long, I'll skip it.

abique:
I've change the PR to target the branch next.
Please, always target the branch next :+1:

include/clap/ext/thread-check.h
tim-janik: >     * The audio-thread may move from one OS thread to another, including the same OS thread as the main-thread

Thanks, I have added this.

>     * There can be only one main-thread and audio-thread at a time

I don't understand that statement.

* At any given point, there may be one main-thread (unlikely to move between OS threads) and multiple audio-threads (quite likely to move between OS threads from a plugin perspective).
* Also, a simple python script that renders a CLAP echo effect onto a WAV file may use only one thread, for which is_main_thread and is_audio_thread are true simultaneously.

BTW, several concurrent data structure implementations make use of caller-provided per-thread tokens. For that, knowing which thread contexts may move between OS threads is crucial, I hope the newest wording is clear enough in that regard.

abique: Yes the host can have a thread pool with thousand of threads, but the host can't call process concurrently on the same plugin instance.
So from the plugin perspective, there can be only one audio thread at a time.

```
void process_plugin()
{
   this->audio_thread_id = std::this_thread::get_id();
   // ...
   this->plug->process(this->plug, ...);
   // ...
   this->audio_thread_id = {};
}
```

PR #138, opened 2022-07-12
Update comments of params.h

xris1658:
"automataions -> modulations" of comments describing modulatable flags.


abique:
Thank you :+1: 
I've changed your PR to target the next branch.

include/clap/ext/params.h
```diff
@@ -129,13 +129,13 @@ enum {
    // Does this parameter support per note automations?
```
baconpaul, line 129: We should get note id comment also. Thanks!

abique: done.

PR #139, opened 2022-07-12
1.1.0

abique:


include/clap/ext/draft/state-context.h
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include "../../plugin.h"
+#include "../../stream.h"
+
+/// @page state-context extension
+/// @brief extended state handling
+///
+/// This extension let the host save and load the plugin state for special purpose.
```
robbert-vdh, line 9:
```suggestion
/// This extension lets the host save and load the plugin state with different semantics depending on the context.
```

include/clap/ext/draft/state-context.h
```diff
@@ -0,0 +1,61 @@
...
+/// Here is a conceptual implementation:
+/// state clap_plugin_state_context.save(ctx) {
+///    state = clap_plugin_state.save();
+///    return reduce(state, ctx);
+/// }
+///
+/// bool clap_plugin_state_context.load(state, ctx) {
+///    state = reduce(state, ctx);
+///    return clap_plugin_state.load(state);
+/// }
```
robbert-vdh, line 25: Is this example necessary? I'm afraid a lot of people will not understand what this means without some additional context (heh). Just saying that the host may freely mix saving and loading operations from the state-context and state extensions and that the plugin should anticipate on that should be sufficient.

robbert-vdh (changes requested):

abique:
OK, I've reverted the latest change as I wasn't sure about it.
Let's go :-)

Bremmers (approved):

baconpaul (approved):
Looks good to me!

abique:
> My only question which I mentioned on discord: Do we want to put the menu extension in here in draft so we can try and implement it in bitwig/surge?

Is it ready? I remember that it still had open spec questions and there is no PR yet for it.
I think we can move on with 1.1.0 and as soon as the menu one is ready we make 1.2.0, unless the menu ext can be ready in the next few days?

tim-janik:
> > My only question which I mentioned on discord: Do we want to put the menu extension in here in draft so we can try and implement it in bitwig/surge?
> 
> Is it ready? I remember that it still had open spec questions and there is no PR yet for it. I think we can move on with 1.1.0 and as soon as the menu one is ready we make 1.2.0, unless the menu ext can be ready in the next few days?

I would recommend to wait and leave specs in draft/ until at least two implementations exist.

Fast iterations are great in the pre-1.0 aera or in draft state but not once things are official API/ABI and should not be deprecated/replaced for a couple years.

It is really easy for an implementation to support an additional "clap.foo.draft/99" interface string next to "clap.foo", but really hard to fix up an interface that lacks a tiny slot for an additional integer as state_context_t has just demonstrated.

PR #148, opened 2022-08-10
Next updates

baconpaul:
This PR

1: Catches next up with 1.1.1
2: Adds a commit to clarify the `const char**` argument in gui preferred api.


abique:
Thank you :+1:

PR #149, opened 2022-08-13
File reference docs

tim-janik:
Provide a (currently missing) description for save_resources() based on discord discussions from last June.

### DIFF
--- a/include/clap/ext/draft/file-reference.h
+++ b/include/clap/ext/draft/file-reference.h
@@ -31,6 +31,10 @@ extern "C" {
 // This describes a file currently used by the plugin
 typedef struct clap_file_reference {
    clap_id resource_id;
+
+   // Flag indicating that the plugin may be able to (re-)install a collection that provides
+   // this resource. DAWs can provide a user option to ignore or include this resource during
+   // "collect and save".
    bool    belongs_to_plugin_collection;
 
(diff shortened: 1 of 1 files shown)


tim-janik:
FYI, [Anklang trunk](https://github.com/tim-janik/anklang/commit/031c81f2efe9dce9a04a761e8c51b0ee4c2d281e) now implements CLAP_EXT_FILE_REFERENCE from the host side.

The docs that I was missing are contained in this pull request. Additionally, I think the docs should specify whether `resource_id` is expected to persist save_state+load_state, i.e. if it's a proper ID based on which the host can resolve missing files after load_state() for plugins that do not support get_blake3_digest().

The method get_file_size() is currently unused, I fail to see a good use for it… Also, the host functions changed() and set_dirty() are mere dummies in my implementation, because files are only ever collected upon save_state() and that first calls save_resources() unconditionally. I'd be happy to get input on these functions and learn what host scenarios could actually benefit from them.

About `belongs_to_plugin_collection`, it took me a while to figure why it's called **plugin**-collection. A DAW could of course detect **host**-collection files via path name checks. Maybe that info could be worth adding to the docs?

abique:
Hi @tim-janik,

Thank you very much for the improvements! :+1: 

Cheers,
Alex

Issue #153, opened 2022-08-31
Missing calling conventions for the function pointers


robbert-vdh:
After pulling my hair out trying to figure out why some plugins would initialize just fine when calling `clap_entry->init()` in this new Wine CLAP plugin bridge I've been working on while others don't, I finally realized that CLAP doesn't specify any calling conventions. Which means that on 32-bit Windows calling any CLAP functions is currently undefined behavior. This is mostly relevant on 32-bit Windows as 64-bit Windows uses a single calling convention, but having explicit calling conventions is still useful as there already are 32-bit Windows CLAP hosts and plugins. Wine these calling conventions are also mandatory when working with Winelib applications like I was.

The fix for this would be to define something like  a `CLAP_ABI` macro that's set to `__cdecl` on Windows and and that's empty on other platforms. All function pointers `bool(*foo)(int bar)` in the CLAP API then need to be changed to `bool(CLAP_ABI *foo)(int bar)`.

EDIT: Fixed some errors, I misremembered some things.

robbert-vdh:
I just realized that the `__stdcall` versus `__cdecl` story is only relevant for 32-bit Windows. But I'd still appreciate a `CLAP_ABI` in CLAP nonetheless, as it's also required for Wine bridging to work. I'll submit a PR for this soon.

sagamusix:
For reference, cdecl usage was also discussed in #21 (but nothing came out of it as the other changes were dropped)

PR #154, opened 2022-08-31
Add explicit calling conventions to all function pointers

robbert-vdh:
This addresses the problem described in #153. Without this systems with multiple calling conventions (like x86 Windows, or a future platform) may not use the correct calling convention when calling these functions. And even if the correct calling convention is used, calling these functions would still be undefined behavior as the default calling convention depends on the compiler. It's also required for being able to use CLAP in a [Winelib](https://wiki.winehq.org/Winelib) setup, as the `__cdecl` calling convention tells GCC to use the Microsoft x64 ABI for those functions.

### DIFF
--- a/include/clap/entry.h
+++ b/include/clap/entry.h
@@ -48,16 +48,16 @@ typedef struct clap_plugin_entry {
    //
    // If init() returns false, then the host must not call deinit() nor any other clap
    // related symbols from the DSO.
-   bool (*init)(const char *plugin_path);
+   bool(CLAP_ABI *init)(const char *plugin_path);
 
    // No more calls into the DSO must be made after calling deinit().
-   void (*deinit)(void);
+   void(CLAP_ABI *deinit)(void);
 
    // Get the pointer to a factory. See plugin-factory.h for an example.
    //
    // Returns null if the factory is not provided.
    // The returned pointer must *not* be freed by the caller.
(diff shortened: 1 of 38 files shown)


baconpaul:
So this changes

```c
int64_t (*read)
```

to

```c
int64_t(CLAP_ABI *read)
```

losing the space after `int64_t`

Can we get the space back?

robbert-vdh:
That's part of the highly customized [.clang-format](https://github.com/free-audio/clap/blob/main/.clang-format) configuration added by @abique. I just reformatted everything (except for the couple of other things in the repo that aren't correctly formatted right now) according to those rules.

robbert-vdh:
I googled for a bit and I couldn't find a way to customize this in clang-format without adding spaces before _every_  paren (which is definitely not what we want here). Maybe it's just an oversight in clang-format?

baconpaul:
> That's part of the highly customized [.clang-format](https://github.com/free-audio/clap/blob/main/.clang-format) configuration added by @abique. I just reformatted everything (except for the couple of other things in the repo that aren't correctly formatted right now) according to those rules.

Ahh right because this change made it look like a function name with typed argument as opposed to a function pointer. Harumph.

abique:
Yo,
Thank you very much for picking this up!
If you need to improve `.clang-format` feel free to do it :-)
Cheers,
Alex

PR #157, opened 2022-09-08
Clarify that input buffer is read-only

jaxter184:
A one-line addition to the documentation intended to codify the consensus in: https://github.com/free-audio/clap/discussions/152


abique:
Thanks!

PR #180, opened 2022-09-28
Clarify Cookie Optionality

baconpaul:
Clarify Cookie Optionality. The documentation was ambiguous as to whether hosts can choose to ignore the plugin cookie. This change documents the fact that they can, and the plugin must react accordingly

### DIFF
--- a/include/clap/ext/params.h
+++ b/include/clap/ext/params.h
@@ -155,16 +155,33 @@ typedef struct clap_param_info {
 
    clap_param_info_flags flags;
 
-   // This value is optional and set by the plugin.
-   // Its purpose is to provide a fast access to the plugin parameter:
+   // This value is optional and set by the plugin. The host will
+   // set it on all subsequent events regarding this param_id 
+   // or set the cookie or nullptr if the host chooses to
+   // not implement cookies. It is very strongly recommended
+   // that the host implement cookies and some plugins may ignore
+   // events which disregard their cookie.
    //
+   // Its purpose is to provide a fast access to the plugin parameter
+   // objects. For instance:
+   //
+   // in clap_plugin_params.get_info
    //    Parameter *p = findParameter(param_id);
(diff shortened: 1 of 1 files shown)


baconpaul:
Just FYI when you read this in the morning @abique - since a lot went by on discord

1. I think most of us think this is wrong and cookies should be host mandatory but
2. If they are not mandatory (which most of us think they should be) then this is an improvement and
3. Either 1 or 2 is better than the current ambiguous state

To make surge meet this optional spec makes a bit of a mess in the way juce wraps a clap and would be super inefficient but I made it work.

abique:
I think the host should be encouraged to provide the cookie if it can do better than N hash map lookup, with N being the number of events addressing parameters.
Otherwise there is no benefit in passing the cookie, the plugin may as well do the resolution.

In Bitwig, once we did scan the parameters, we build an array with all the cached parameters info and internally work with param index, so we get immediate access to the cookie, therefore the optimization works.

abique:
I'm 100% convinced that the cookie must not be mandatory and the explanation is short:
- there can be some scenario where the cookie can't be provided
- param_id is enough to resolve the parameter

The scenario is a device in which the user would manually enter the GUI a parameter id and amplitude, and then the device would generate parameter modulation events which would be merged into the event stream. This device can't know the cookie, and the host providing the event stream may not be aware of plugin parameters.

CLAP needs to be working for all, for a very long time, and prove to be flexible enough to work in situations we did not even think of. Making the cookie mandatory reduces the flexibility, and may make it inappropriate for certain things.

baconpaul:
I have updated the juce wrappers and also updated surge to support null cookies.

abique:
I'll merge this on Friday unless anyone comes with strong arguments against it.

baconpaul:
I mean you know my arguments against it :)

But we have to pick one. So that's fine with me.

PR #193, opened 2022-10-18
resource directory extension

abique:


abique:
Generated pkgconfig file looks good:
```
prefix=/usr/local
includedir=${prefix}/include

Name: clap
Description: The interface headers for the CLAP audio plugin API
Version: 1.1.3

Cflags: -I${includedir}
```

PR #196, opened 2022-10-31
fix: make features on plugin_descriptor_t const

mnkisala:
I'm not sure if there would be a reason for anything to modify the list of features after instantiating the descriptor (which should take place at compile-time anyway). I was trying to port [plugin-template](https://github.com/free-audio/clap/blob/main/src/plugin-template.c) over to Zig and the compiler wouldn't let me cast __*const [*c]const u8__ to __[*c][*c]const u8__ (obviously that would discard the const, which is allowed in C but not in Zig).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,82 +1,91 @@
-# Changes in 1.1.2
+Changes in 1.1.2
+================
 
-* [macros.h](include/clap/private/macros.h): add `CLAP_ABI` which defines the calling convention and use it everywhere
-* [events.h](include/clap/events.h): clarify `clap_input_events.size()`
-* [check-for-update.h](include/clap/ext/draft/check-for-update.h): fix typedef
-* [file-reference.h](include/clap/ext/draft/file-reference.h): improve documentation
-* [params.h](include/clap/ext/params.h): clarify how the cookie works and add some notes about `flush()`
-* [process.h](include/clap/process.h): clarify how the audio buffer mapping works
-* [gui.h](include/clap/ext/gui.h): clarify `clap_plugin_gui.get_preferred_agi()`
-* [plugin-factory.h](include/clap/plugin-factory.h): mention `clap_plugin_invalidation_factory` which can be use to invalidate cached entries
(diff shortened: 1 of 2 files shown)


baconpaul:
won't this be an ABI breaking change though?

robbert-vdh:
It technically would be, although realistically no host would modify this array. So in practical terms this would only break the API and not the ABI.

baconpaul:
there's not a mangled name change or anything like that? i guess this is c code and not a function or top level global point so the mangler shouldn't be in effect.

robbert-vdh:
Yes it's just a struct field. The memory layout for a `const foo**` and a `const foo* const*` is identical since both are just a pointer.  Of course, it's still technically an ABI change because the older version would let you overwrite the feature string pointers while the newer version does not, but in practical terms this won't break anything and this PR's version is more in line with the intended semantics.

robbert-vdh:
On another note, this PR should be targeting the `next` branch. Not `main`.

abique:
Please include the changelog entry and then it'll be ready to merge I suppose.

ChangeLog.md
```diff
@@ -1,82 +1,91 @@
...
-* [midi-mappings.h](include/clap/ext/draft/midi-mappings.h)
-* [preset-load.h](include/clap/ext/draft/preset-load.h)
-* [quick-controls.h](include/clap/ext/draft/quick-controls.h)
-* [surround.h](include/clap/ext/draft/surround.h)
-* [track-info.h](include/clap/ext/draft/track-info.h)
-* [transport-control.h](include/clap/ext/draft/transport-control.h)
-* [tuning.h](include/clap/ext/draft/tuning.h)
-* [voice-info.h](include/clap/ext/draft/voice-info.h)
+Changes in 1.1.2
+================
```
robbert-vdh, line 2: You may want to disable your auto formatter heh.

robbert-vdh (changes requested):

abique:
Thanks!

PR #204, opened 2022-11-05
Win: Correct non-existing environment variable and native directory separator

Schroedingers-Cat:
This PR improves two things in the `entry.h` file's comment section:
- On Windows, the system-wide default deployment path for CLAP plugins is `C:\Program Files\Common Files\CLAP`, which is stored in the environment variable `%COMMONPROGRAMFILES%`. However, the comment section in `entry.h` specifies a wrong (non-existing) environment variable to be used.  
- The directory separator on Windows is `\` but the comment section uses the *nix directory separator `/` for Windows paths.

### DIFF
--- a/include/clap/entry.h
+++ b/include/clap/entry.h
@@ -16,8 +16,8 @@ extern "C" {
 //   - /usr/lib/clap
 //
 // Windows
-//   - %CommonFilesFolder%/CLAP/
-//   - %LOCALAPPDATA%/Programs/Common/CLAP/
+//   - %COMMONPROGRAMFILES%\CLAP
+//   - %LOCALAPPDATA%\Programs\Common\CLAP
 //
 // MacOS
 //   - /Library/Audio/Plug-Ins/CLAP


abique:
So it seems that %CommonFilesFolder% is for MSI installer: https://learn.microsoft.com/en-us/windows/win32/msi/commonfilesfolder

abique:
Windows accepts both `/` and `\` as directory separator isn't it?

abique (approved):

abique:
https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables

mkruselj:
PR looks good to me! Windows does seem to accept both directory separators, but it's still using \ pretty much everywhere throughout its interface.

abique:
Could it be that we provide both the MSI installer variables and environment variable in the documentation? Or would it be misleading?

mkruselj:
Not everyone would be writing a MSI installer (there's InnoSetup, InstallShield, NSIS...) so I think it's better to use the system envvar here, it's super clear then and developers can consult their installer's documentation for additional clarification (if they already don't know these things by heart).

abique:
Then we're good to go I believe.

defiantnerd (approved):
Indeed the path separator is usually backward slash, although nearly all APIs accept forward slashes, too. 

All uppercase for environment variables is also standard.

I think this improves the clarity of the documentation.

robbert-vdh:
> Windows accepts both `/` and `\` as directory separator isn't it?

Pretty much everywhere except for inside of the GUI file open dialog. So for cross-platform code it can be a good idea to just use forward slashes everywhere to reduce the amount of platform-specific bits mixed in with the rest of the code.

PR #219, opened 2022-11-28
Work on channel-info to replace track-info

abique:


include/clap/ext/draft/channel-info.h
```diff
@@ -0,0 +1,61 @@
...
+   CLAP_CHANNEL_INFO_HAS_TRACK_COLOR = (1 << 1),
+   CLAP_CHANNEL_INFO_HAS_TRACK_ID = (1 << 1),
+   CLAP_CHANNEL_INFO_HAS_TRACK_PATH = (1 << 1),
+   CLAP_CHANNEL_INFO_HAS_CHANNEL_NAME = (1 << 2),
+   CLAP_CHANNEL_INFO_HAS_CHANNEL_COLOR = (1 << 3),
+   CLAP_CHANNEL_INFO_HAS_CHANNEL_ID = (1 << 4),
+   CLAP_CHANNEL_INFO_HAS_CHANNEL_PATH = (1 << 4),
+   CLAP_CHANNEL_INFO_HAS_AUDIO_CHANNEL_COUNT = (1 << 5),
+   CLAP_CHANNEL_INFO_HAS_AUDIO_PORT_TYPE = (1 << 6),
+};
```
robbert-vdh, line 24: Minor thing, but it's probably a good idea to have this follow the same order as the struct field.

include/clap/ext/draft/channel-info.h
```diff
@@ -0,0 +1,61 @@
...
+
+typedef struct clap_channel_info {
+   uint32_t flags; // see the flags above
+
+   // info relatives to the track to which the current channel belong
+   char         track_id[CLAP_CUSTOM_ID_SIZE];
+   char         track_name[CLAP_NAME_SIZE];
+   char         track_path[CLAP_PATH_SIZE];
+   clap_color_t track_color;
+   bool         is_return_track;
```
robbert-vdh, line 34: Before this is merged, every field needs a very clear explanation of what the expected values are. And also mention things like, if both the track and channel path are set, then the track path must be a prefix of the channel path. And I'm not sure if 'channel' is the clearest name for this. Not sure what a better alternative would be though. Sub-track? 'Plugin path' also made sense to me, except for the part where the path points not to the plugin but to the thing holding the plugin so that's also not a good choice.

robbert-vdh (changes requested):

include/clap/ext/draft/channel-info.h
abique: Maybe I should rename it to `device_chain_info`.

include/clap/ext/draft/track-info.h
```diff
@@ -4,24 +4,76 @@
...
+//
+// The terminology used in this extension:
+// A track is a channel.
+// A channel contains one device chain and usually a mixer at the end.
+// A device chain, contains a bunch of devices.
+// A device may contain nested channels, return channels (for send FX) and device chains.
+//
+// For example, consider a Drum Machine device, each drum pads contains a channel which contains
+// themselves device chains. Additionnaly a Drum Machine device can contain return channels for
+// send FX.
```
robbert-vdh, line 19: Couple typos here, but I guess we can fix docs before this is merged into the master branch.

include/clap/ext/draft/track-info.h
```diff
@@ -4,24 +4,76 @@
 #include "../../color.h"
 #include "../../string-sizes.h"
 
+// This extensions let the plugin query info about the track and device chain it's in.
+// It is useful when the plugin is created, to initialize some parameters (mix, dry, wet)
+// and pick a suitable configuartion regarding audio port type and channel count.
+//
+// The terminology used in this extension:
+// A track is a channel.
+// A channel contains one device chain and usually a mixer at the end.
```
robbert-vdh, line 13: Maybe it's a idea to use the word 'track' here instead of channel? Especially since channel also refers to audio channels.

robbert-vdh (approved):

include/clap/ext/draft/track-info.h
himwho: +1 on reserving "channel" for referring only to a channel of audio in an audio stream for cases of multichannel processing

robbert-vdh:
Not quite sure what plugins would do with it, but would it also make sense to add another field with flags for the track/device chain state? So soloed, muted, etc.

abique:
> Not quite sure what plugins would do with it, but would it also make sense to add another field with flags for the track/device chain state? So soloed, muted, etc.

Yes, solo, muted, record can be useful.
But if the processing depends on that, then you need to make it available on the audio-thread too.

include/clap/ext/draft/track-info.h
abique: How would you call the return channel inside a drum machine?

robbert-vdh:
Those things could also be added to `clap_transport_flags`.

include/clap/ext/draft/track-info.h
robbert-vdh: `is_device_chain_return` would be consistent with the rest of the names.

Bremmers:
Plugins might want to know they're in the master section. A compressor or EQ can then default to settings suitable for mastering, for example.

robbert-vdh:
Good idea. That could perhaps also be part of that bit field. And would it also make sense to move the return track fields there?

Bremmers:
Perhaps the return track field should be renamed to "prefer wet-only". That's the only purpose for it I think?

There can be situations where wet-only is preferred, but there isn't really a 'return channel'.

abique:
It seems there are two possible approaches:
1. Describe the environment, which implies what the plugin should do
  - send channel -> wet 100%
  - master track -> master preset on init
  - group track -> bus processing on init
2. Tell the plugin what to do directly
  - wet 100%
  - for_master_processing
  - for_bus_processing
  
Maybe 2. is a better call for consistent behavior.

abique:
Also I think that describing the environment is maybe too much work for this extension:
- can be inside a device
- device can have master section and return channel
- device can have muted/solo pads
- device chain can be inside a device, or a track
- track can be inside track group, ...

So describing correctly the whole document / path will be a tough job, and either we do it well or we don't do it.

abique:
I'm quite happy with this extension now.

PR #220, opened 2022-11-29
Audio port activation

abique:


robbert-vdh (changes requested):
The `CLAP_EXT_AUDIO_PORTS_ACTIVATION` constant should include `/draft0`, but otherwise LGTM.

include/clap/ext/draft/audio-ports-activation.h
```diff
@@ -31,6 +32,10 @@ extern "C" {
 #endif
 
 typedef struct clap_plugin_audio_ports_activation {
+   // returns true if the plugin supports activation/deactivation while processing.
+   // [main-thread]
+   bool(CLAP_ABI *can_activate_while_processing)(const clap_plugin_t *plugin);
+
    // activate the given port
    // [main-thread]
```
robbert-vdh, line 40: Oh, if calling `set_active()` while processing is allowed (which it should be if you don't want xruns/plugins dropping out for a brief moment/plugins resetting every time IO is connected), then this should probably be called on the audio thread, right? Otherwise you need a bunch of blocking synchronization or the flag may change in the middle of a process call.

robbert-vdh (changes requested):

include/clap/ext/draft/audio-ports-activation.h
abique: In Bitwig we'd check before activating, and if the plugin doesn't support it, we may keep the port activated all the time.

defiantnerd: If a host does not optimize this and keeps all ports "connected", it would be okay.
The default state should be "activated" to make a smooth transition for plugins?

abique: Port (de)activation while the plugin is processing is a consequence of a user action, like a mouse click.
So it happens on the main-thread, and isn't a real-time thing with a sample accurate rendering.

When loading the plugin, port deactivation should be done before activating the plugin in order to have a correct and reproduce-able rendering.

Regardless of processing state, this method should be called on the main thread and the plugin should manage to communicate that change to its DSP. Again it isn't important to be "sample accurate" here, this whole thing is just to be able to activate a port without interrupting the audio, in a best effort.

I don't like having one function which depending on the state should be called either on the audio-thread or main-thread, this makes the interface confusing, and can be racy. Also it could be that the plugin needs to adjust a few things on its GUI so it'd also need the info on the main thread anyway.

robbert-vdh: But lots of user interactions in DAWs trigger effects that happen on the audio thread. Notifying the plugin that the port is active or not may affect the plugin's DSP (which is the entire point of this extension). It's impossible for the plugin to synchronize this correctly unless the host guarantees that this function is called in between the process call where the audio port was deactivated and the next process call where the audio port is activated.  The host is the only one that can do the proper synchronization, because it knows the exact moment when the connected port becomes part of the audio graph. It would just call this function before the process call (or it could even be an event to avoid a second audio thread function call). Then the plugin always knows whether a port is connected or not at the start of a process call.

Bremmers: Doing it in the main thread works fine for the use case Alexandre has in mind. But doing it in the audio thread may be more future proof (when people may want to use this for other scenarios). Using an event for this, as per Robbert's suggestion, could be a nice solution.

abique: There is one problematic scenario with doing it on the audio thread:
1. user connects input in the host gui
2. host performs all the document routing updates (main thread)
3. host queues changes to the audio engine
4. sound card is disconnected
5. bus activation event for the plugin is lost (was in the queue)

While if the plugin bus activation is done in 2., then it can't be lost.
Also there is no point in begin sample perfect here, we're just trying to be able to (de)activate without restarting, plugin can just set a flag and deals with this state change itself.

abique: Also to me having it always on the main thread is superior because it kills the question "on which thread am I?".

And pro-plugins will probably have some GUI changes when a bus is (de)activated so they'll need that information on the main-thread, so if we were passing it on the audio thread, they would then have to communicate it back to their GUI from the audio-thread.

To me having a consistent entry point and always pass the info on the main thread is superior.

abique: Let's continue this discussion in the "next" PR

PR #221, opened 2022-11-29
Context menu

abique:


nakst:
I propose that the `label` parameter on the `add_entry` and `add_checkbox` functions should have a special behaviour when it contains the character `&`.

If the GUI library the menu is to be displayed in has some sort of access key functionality (see https://learn.microsoft.com/en-us/windows/win32/uxguide/inter-keyboard#menu-access-keys), then the letter (only uppercase and lowercase Latin A-Z are allowed) after the `&` should be the access key, and `&` should be hidden.

If the GUI library does not have access key functionality, the `&` should be ignored and hidden.

Trinitou:
As this is similar to #222 as @robbert-vdh mentioned (thank you!) -> Is this context menu extension about context menus for inidividual parameters or also for the whole plugin instance? Maybe it might be helpful to clarify that.

robbert-vdh:
Yes, it's for parameters. You can think of this extension as a streamlined version of VST3's context menu interface.

Trinitou:
> Yes, it's for parameters. You can think of this extension as a streamlined version of VST3's context menu interface.

Maybe then it should somehow be named accordingly. Something like "param-context-menu" ?

abique:
> > Yes, it's for parameters. You can think of this extension as a streamlined version of VST3's context menu interface.
> 
> Maybe then it should be named accordingly. Something like "param-context-menu" ?

You can have entries unrelated to parameter if you pass `CLAP_INVALID_ID`.

robbert-vdh:
> I propose that the `label` parameter on the `add_entry` and `add_checkbox` functions should have a special behaviour when it contains the character `&`.
> 
> If the GUI library the menu is to be displayed in has some sort of access key functionality (see https://learn.microsoft.com/en-us/windows/win32/uxguide/inter-keyboard#menu-access-keys), then the letter (only uppercase and lowercase Latin A-Z are allowed) after the `&` should be the access key, and `&` should be hidden.
> 
> If the GUI library does not have access key functionality, the `&` should be ignored and hidden.

I had already proposed this elsewhere, but my preference would be to move one letter accelerators to a second argument that would take a single character case insensitive string. If the host doesn't support accelerators, then it wouldn't need to do anything except for maybe escaping the string to match their GUI toolkit's context menu label format (which they'd need to do anyways). If the host does support accelerators, it would use that letter as an accelerator (and perhaps underline the first occurrence of that letter if that's how they indicate accelerators). That way the label string doesn't contain magic symbols, doesn't need to be parsed for those magic symbols (to either remove or replace them), and isn't tied to the format expected by a single GUI toolkit (Win32 controls in the case of `&`).

abique:
Hi,
I've improved this extension to have a concept of target, so for now it is either global or param, but we could think of having different targets in the future.
I've also added a accelerator description, which allows for key combo too (up to four).

abique:
The description will be improve but first let's see if the design is clear and if we like it.

robbert-vdh:
The current version supports both modifier keys and also entire key chords for context menu accelerators. Do we want this? Is there any mainstream GUI toolkit that supports this? Or is this functionally intended as part of extending context menus into a full generic action API? The way it's currently set up makes the extension quite a bit more complex than the initial draft was.

abique:
I think vscode has it so I though key combo was a common thing now.

nakst:
As I said in my previous comment, I think there should be a way to specify the access key for a menu item. As robbert-vdh has argued, this is probably best done by adding an argument `char access_key` to `add_entry` and `add_checkbox`.

Also, to address robbert-vdh's concerns about inclusion of keyboard accelerators, I think the CLAP documentation should make it clear that the accelerator parameter in this API does *not* create any form of binding, but only exists as a decoration for the menu item.

robbert-vdh:
> I think vscode has it so I though key combo was a common thing now. If one doesn't have accelerator, it is possible to pass `NULL`.

Just opened VsCode, the keybindings it shows there are global keybindings that can always be used, even when the context menu is not open. So it's just a reminder of how to access that action with a keyboard shortcut. In fact, VsCode doesn't seem to use single letter accelerator keys in their context menus at all. 

You also can't use those shortcuts _while_ the context menu is open. They really are just reminders of the global key bindings. So my question still is, are these supposed to be global shortcuts that are always active even when no context menu is visible? If so, how do you handle clashes with the DAWs own shortcuts?

abique:
> > I think vscode has it so I though key combo was a common thing now. If one doesn't have accelerator, it is possible to pass `NULL`.
> 
> Just opened VsCode, the keybindings it shows there are global keybindings that can always be used, even when the context menu is not open. So it's just a reminder of how to access that action with a keyboard shortcut. In fact, VsCode doesn't seem to use single letter accelerator keys in their context menus at all.
> 
> You also can't use those shortcuts _while_ the context menu is open. They really are just reminders of the global key bindings. So my question still is, are these supposed to be global shortcuts that are always active even when no context menu is visible? If so, how do you handle clashes with the DAWs own shortcuts?

Usually they are just reminders, so you can learn the shortcuts while reading the menu.

abique:
> I think the CLAP documentation should make it clear that the accelerator parameter in this API does _not_ create any form of binding, but only exists as a decoration for the menu item.

Yes

robbert-vdh:
> Usually they are just reminders, so you can learn the shortcuts while reading the menu.

Exactly. They're super useful for that, but because they're reminders of global hotkeys there's also no reason for them to be in this extension. Unless this extension is also supposed to register global hotkeys on the host.

abique:
> > Usually they are just reminders, so you can learn the shortcuts while reading the menu.
> 
> Exactly. They're super useful for that, but because they're reminders of global hotkeys there's also no reason for them to be in this extension. Unless this extension is also supposed to register global hotkeys on the host.

I agree, they are not essential.

baconpaul:
Sorry to have missed most of this conversation

The build construct is super clever. I really like it. Easy to implement also.

The condition in which the host calls the plugin with a builder (as opposed to vice versa) is so the plugin can add items to the hosts plugin menu? That's awesome!! I will definitely use that.

the accelerator stuff is tricky I agree but looks like a good implementation here. I could definitely implement the surge features we have if this api was in place.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+// There can be different target kind for a context menu
+enum {
+   CLAP_CONTEXT_MENU_TARGET_KIND_GLOBAL,
+   CLAP_CONTEXT_MENU_TARGET_KIND_PARAM,
+};
+
+// Describes the context menu target
+typedef struct clap_context_menu_target
+{
+   uint32_t kind;
```
Trinitou, line 23: Maybe better call it 'type' ?

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+   bool(CLAP_ABI *populate)(const clap_host_t                 *host,
+                            const clap_context_menu_target_t  *target,
+                            const clap_context_menu_builder_t *builder);
+
+   // If target is null, assume global context
+   // [main-thread]
+   bool(CLAP_ABI *perform)(const clap_host_t                *host,
+                           const clap_context_menu_target_t *target,
+                           clap_id action_id);
+
```
Trinitou, line 103: Just for my understanding:

- If I understand correctly, calling populate and perform on the host from the plugin is for showing the plugin's own context menu but with (additional) host entries
- But through the clap_host_context_menu::popup method, it also could be the host's popup menu that shows up on top of the plugin. So clap_host_context_menu::populate and clap_host_context_menu::perform are a somehow redundant, aren't they?
- Or is it just so that we can have both options for context menus on top of the plugin?

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+                             clap_id                                 action_id);
+
+   // accelerator may be null
+   bool(CLAP_ABI *add_checkbox)(const struct clap_context_menu_builder *builder,
+                                const char                             *label,
+                                const clap_context_menu_accelerator_t  *accelerator,
+                                bool                                    is_enabled,
+                                bool                                    is_checked,
+                                clap_id                                 action_id);
+
```
Trinitou, line 67: Maybe there could be way to handle redundant menu entries. E.g. if both the host and the plugin want to add a 'reset to default value' menu entry? So I think it might be a useful addition if add_entry and add_checkbox had an additional parameter which is an enum value for standard menu entries. This way the menu builder can omit redundant entries.
Useful enum values might be CUSTOM (for non-standard), RESET_TO_DEFAULT, SET_TO_MIN, SET_TO_MAX, COPY_VALUE, PASTE_VALUE, COPY_PARAM_ID_TO_CLIPBOARD, ...
If we want to go totally crazy, even submenus might be standardized (like a common 'add modulation from source' menu). So there could be a second enum for standard submenus and an additional parameter for begin_submenu.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+static CLAP_CONSTEXPR const char CLAP_EXT_CONTEXT_MENU[] = "clap.context-menu.draft/0";
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// There can be different target kind for a context menu
+enum {
+   CLAP_CONTEXT_MENU_TARGET_KIND_GLOBAL,
+   CLAP_CONTEXT_MENU_TARGET_KIND_PARAM,
```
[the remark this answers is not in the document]
abique: What is your question here?

include/clap/ext/draft/context-menu.h
abique: @baconpaul you're native speaker, what fits better here type or kind?

include/clap/ext/draft/context-menu.h
abique: populate will add the host entries into the menu, so then the plugin can show its own menu, including the host entries.

perform is for executing an action after the menu item has been clicked.

If you call `host->popup()`, then the host will call `plugin->populate()` then show the menu then if a plugin menu item is clicked `plugin->perform()`

include/clap/ext/draft/context-menu.h
abique: I think it is too specific.
Maybe by convention, plugin should only inject action that the host can't do, and vice-versa.

If the host is showing the menu and asking the plugin to populate its items, then all the operations like reset to defaults, etc... should already be provided by the host. In that case the plugin should only provide menu entries that complement what the host could achieve without this menu extension.

include/clap/ext/draft/context-menu.h
baconpaul: English language either works. Have we made a tyoe / kind choice elsewhere in extensions? That’s what I would look for

include/clap/ext/draft/context-menu.h
baconpaul: So we have this problem in surge where both surge and reaper have a midi learn option and both appear on the menu. They do different things

I agree that enumerating functions is way too specific - as evidenced by the one where we have a problem not being on your list :) - and plugins just deal with this using the information provided about the host and items if it is a problem

include/clap/ext/draft/context-menu.h
robbert-vdh: Type is a reserved keyword in many languages, but CLAP already uses the word in other places as well. (IIRC the Rust bindings use either `type_` or `typ` for those fields/params, don't recall which one is used)

robbert-vdh:
Yeah I strongly feel like the complex keyboard shortcuts in the current version of this PR should be replaced by a simple single character accelerator string. Complex keyboard shortcuts like this cannot be used when the context menu is open (so they'd be dead on arrival), and in other applications that do list keyboard shortcuts in context menus they're reminders of global shortcuts, which aren't a thing in this case.

Bremmers:
Well, having global shortcuts is not entirely impossible. Parameter menus could have an 'Set all params to default' item for example. And there can be global menus that aren't tied to a param, plugins could use these for anything.

I agree the Windows '&' accelerators should be supported too.

Bremmers:
On second thought I think we don't need to do anything for the Windows '&' accelerators: the side that shows the menu can just assign a suitable character itself. This avoids convoluting the API, and there can't be conflicts between the host and plugin menu item hotkeys. And it's easy to do. Delphi menus, for example, can even assign the hotkeys automatically.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+// Describes a single key strike with its modifiers
+typedef struct clap_context_menu_accelerator_key {
+   uint32_t flags;
+   uint32_t key; // ASCII key
+} clap_context_menu_accelerator_key_t;
+
+// Describes an accelerator which can consist of a key strike combo
+typedef struct clap_context_menu_accelerator {
+   uint32_t combo_size;
+   clap_context_menu_accelerator_key_t keys[4];
```
Bremmers, line 46: Why keys[4]? Just one would do I think.

robbert-vdh: Key chords. But, I still don't think any of this is useful or even usable for an extension like this. An `accelerator` parameter that accepts a single character string should suffice.

robbert-vdh: I guess the term chords for this is not very widely used outside of Emacs. What I'm referring to here is a key binding with one or more leader keys. So for instance `Ctrl+C` followed by `Alt+G` followed by `Ctrl+C` followed by `d`, or `C-c M-g C-c d` for short. But, I don't know why context menus would need keybindings like this (and again, keybindings more complex than single letter accelerators aren't supported in most toolkits).

baconpaul: Obligatory xkcd: https://xkcd.com/2583/

I also think it is unlikely a plugin and host could negotiate non overlapping keyboard accelerators. Like if bitwig tells me alt 3 does something and I use it for something else I will ignore bitwig. And if I don’t have keybindgd I will ignore bitwig. The accelerator display is useful perhaps but are you expecting the plugin will actually process the key events!

Bremmers: Key chords... learned something new ;-)

I guess a host doesn't have that many shortcuts of its own while a plugin has focus? It could work, but it's probably overkill indeed.

Trinitou: 1. Idea A: Maybe there should be an extra extension for plugin keyboard accelerators?
I mean, it always has been problematic with overlapping host/plugin accelerators. So if they would communicate about that with each other, they could maybe find a better compromise? Like the plugin requests a set of accelerators or the host can tell the plugin which accelerators are forbidden. Something like this. So I arrive at...

2. Idea B: Or maybe keyboard accelerator concepts would better fit into a "commands"(&settings) extension where the plugin tells the host its set of commands/toggle settings/enum setting, so that the host could set up a plugin-specific key command editor (where it can prevent the user from assigning duplicate accelerators).
(Another argument for this approach would also be that it is unlikely that there are key accelerators for parameter-specific commands/settings but rather only for global commands.)
Also a separate commands/settings extension would allow the host to MIDI-learn plugin commands which would be awesome (but would not really fit into a context menu extension)

Trinitou: Ah yes, my most important point: All in all I think that we maybe even could leave the accelerators completely out here... Or just make it a simple character string as others already proposed here (even if the plugin cannot guarantee that the accelerator will actually work)

robbert-vdh: Re: 1), accelerators are allowed to overlap in pretty much every GUI toolkit. You'd just press the same key to cycle between the options and then hit enter. I personally never use the context menu accelerators at all though.

Trinitou: Oh, I think I mixed up terminology here: My entire comments were about shortcuts/key commands, not accelerators

defiantnerd: In a context menu, there would be key bindings like a simple letter. As example on Windows this is such a thing as "&Cut" where `C` gets underlined, and a keypress to `C` selects the entry. This is called **navigation key** or **hotkey**. They work only when the context menu is open.

Accelerators are usually a keyboard shortcut for a (program global) function in a menu bar above the window.
Here CTRL+C would be the accelerator for "Cut", where the entry is looking like "Cut     CTRL-C".

Since this is a context menu, we talk about hotkeys, right? In this case, modifiers would not be needed, since using a modifier key immediately closes such a menu.

Trinitou: Thank you for clarification, @defiantnerd! Ok, then I really talked about "accelerators". :) Really sorry for confusion! :/

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+
+   // accelerator may be null
+   bool(CLAP_ABI *add_entry)(const struct clap_context_menu_builder *builder,
+                             const char                             *label,
+                             const clap_context_menu_accelerator_t  *accelerator,
+                             bool                                    is_enabled,
+                             clap_id                                 action_id);
+
+   // accelerator may be null
+   bool(CLAP_ABI *add_checkbox)(const struct clap_context_menu_builder *builder,
```
defiantnerd, line 61: Shouldn't this be called `add_checkmark`, `add_checkmark_option` or `add_option` (since it makes a checkmark and not a checkbox)?

include/clap/ext/draft/context-menu.h
defiantnerd: So, should we remove the accelerator part? It is a context menu, opened with right-click, correct?
A different beast would be adding menu entries to a host provided menu bar.

include/clap/ext/draft/context-menu.h
abique: Then better to continue with `kind` I suppose.

include/clap/ext/draft/context-menu.h
abique: Sorry for the delay, last week has been intense for me and I could not find the time to get back to it.

I'd like to clarify one aspect of this extension:
- accelerator are just here to tell the user how the plugin (or host depending on who provides the menu item) is processing its shortcut, it isn't meant to provide a way to exchange accelerators (or shortcuts) between the host and plugins

Then about simplifying the accelerator, if instead we describe them using a string `"&A"` for `Alt+A` how do you then do `Alt+&` is it `"&&"`?
So we could have:
- `#` for Command/Windows
- `&` for Alt
- `@` for Ctrl
- `!` for Shift
- `,` for separating key chords

`"&#k,!A"` for `Alt+Cmd+k, Shift+A`
`"&##"` for `Alt+Cmd+#`

Is that really better?

include/clap/ext/draft/context-menu.h
abique: Yeah maybe, I don't have a strong opinion on that.
To me checkbox is a toggle entry.
Options would be more if you have a set of entries which are exclusive to each other (radio buttons in some toolkits).

include/clap/ext/draft/context-menu.h
abique: To me, using a string would then force to have a parser which will then produce something equivalent to what `clap_context_menu_accelerator_key` is. So I'm far from being convinced here, also when your application becomes more complex, you probably have an internal representation of a shortcut which isn't a string itself.

Here is what most will do:
```C
clap_context_menu_accelerator_key_t accel = convert_to_clap(my_accel);
...
```

The current interface is maybe more work at first, but I suppose that in the long run, there will be less issues, because parsing a string is always a source of bugs.

include/clap/ext/draft/context-menu.h
robbert-vdh: Shouldn't that add a checkbox, and not a check mark?

A checkbox is one of these things: (https://en.wikipedia.org/wiki/Checkbox)

- [ ] Unchecked
- [x] Checked

This is a check mark: :heavy_check_mark: (https://en.wikipedia.org/wiki/Check_mark)

I'd assume this `add_checkbox()` option should add a checkbox that can be used to toggle a boolean value, right?

robbert-vdh: So what builder option do you call when you don't have a check mark to display? The check mark option...but without the check mark? Like you said, other GUI usually call it just a 'check menu item'. So `add_check_entry()` would probably be the most consistent and least confusing option.

Issue #222, opened 2022-11-29
plugin actions extension


Trinitou:
What if a plugin could register actions to the host?

The host could then:
- list the actions in a right click menu of the plugin
- map a hardware controller to those actions
- let the user put trigger buttons on remote control pages (Bitwig-specific)

Use case examples would be
- randomize actions (e.g. for step-sequencer step values)
- learn actions for adaptive audio algoriths (e.g. plugins that listen and adapt to the loudness or spectral content of incoming audio)
- shake action for a spring reverb
- panic action for FX involving audio feedback loops

More a rough idea but I imagine this would be a very nice feature for CLAP
Something similar was already mentioned in #32 but I think putting it into a separate issue might help to bring on the discussion.

nakst:
I second this. It's also an easy way for improving plugin accessibility, if the host has implemented the OS's automation APIs.

robbert-vdh:
What's the difference between this and the proposed context menu extensions? #221 is a new draft based on the earlier discussions.

Trinitou:
> What's the difference between this and the proposed context menu extensions? #221 is a new draft based on the earlier discussions.
Oops, didn't see that. But I think #221 is more about context menus for individual parameters. Correct me if I'm wrong, @abique !
This one is about actions that are available for a whole plugin instance. I'm not sure whether actions like I mentioned as examples should rather be parameters themselves with a parameter id (so that there are "trigger action parameters") or if these actions are some kind of extra type. I suppose somthing like "randomizing a step sequencer" might be a bit more CPU intensive so at least it shouldn't be modulatable nor automatable.
Also this proposal would be independent from any context menu handling as the host could decide by itself whether to put plugin actions into a context menu, have them midi-learnable, and/or map them to a hardware controller.

Trinitou:
> @Trinitou that'd be nice to have trigger port/events for the plugin. Those controls could indeed be mapped to hw controller. I think that would do what you're asking for?

Yes, I think that would do the job. OT have the triggers as parameters would help so that they could be put into remote control pages but maybe that would confuse things (?)

nakst:
I think the context menu API draft could work as-is for this purpose.
When the host wants to query the plugin's 'actions', it can call `clap_plugin_context_menu_t.populate` with some special reserved value for the `param_id` indicating the 'global actions list'. Whether it actually displays the populated items in a menu or in some sort of list for assigning to controllers is up to the host.

nakst:
[...]

PR #229, opened 2022-12-02
Triggers extension 

Trinitou:
Already discussed the idea for an trigger action extension in #222.

I came to the conclusion that a modified version of the params extension might be the best way to go about it so here it is a first proposal.


Trinitou:
Some things I'm not totally sure about yet so input/discussion is very welcome:
- Not totally sure whether the request_flush mechanism is needed for trigger actions.
- Same goes for the cookie mechanism.
- First I thought that sample-accurate action triggering isn't needed as note-on messages might already do the job. But the more I think about it the more I can see that there is a place for sample-accuracy here as well.
  - (Also actions that are too CPU-heavy for sample-accuracy just shouldn't have the CLAP_ACTION_IS_AUTOMATABLE flag set and might happily ignore the sample offset information.)
  - Also in the future, Bitwig modulators could trigger sample-accurate plugin actions (with the modulation value crossing 0.0 or 0.5)
- I don't know yet of any DAW / plugin that has such a trigger action functionality so is this extension already relevant? And even per-voice? Otherwise one has to start. And at least I already can imagine many great use-cases 😃
- As mentioned [here](https://github.com/free-audio/clap/issues/222#issuecomment-1330995728), it might be desirable to have a set of standard action IDs, like e.g. for something like "initialize current patch" ...

include/clap/events.h
```diff
@@ -113,6 +113,10 @@ enum {
    CLAP_EVENT_MIDI,       // raw midi event; clap_event_midi
    CLAP_EVENT_MIDI_SYSEX, // raw midi sysex event; clap_event_midi_sysex
    CLAP_EVENT_MIDI2,      // raw midi 2 event; clap_event_midi2
+
+   // Represents a trigger action
+   // Uses clap_event_action_trigger.
+   CLAP_EVENT_ACTION_TRIGGER,
```
abique, line 119: CLAP_EVENT_TRIGGER

include/clap/events.h
```diff
@@ -259,6 +263,20 @@ typedef struct clap_event_midi2 {
    uint32_t data[4];
 } clap_event_midi2_t;
 
+typedef struct clap_event_action_trigger {
+   clap_event_header_t header;
+
+   // target action
+   clap_id action_id; // @ref clap_action_info.id
```
abique, line 270: trigger_id

include/clap/events.h
```diff
@@ -259,6 +263,20 @@ typedef struct clap_event_midi2 {
    uint32_t data[4];
 } clap_event_midi2_t;
 
+typedef struct clap_event_action_trigger {
```
abique, line 266: clap_event_trigger

include/clap/ext/draft/actions.h
```diff
@@ -0,0 +1,166 @@
+#pragma once
+
+#include "../../plugin.h"
+#include "../string-sizes.h"
+
+static CLAP_CONSTEXPR const char CLAP_EXT_ACTIONS[] = "clap.actions.draft/0";
```
abique, line 6: CLAP_EXT_TRIGGERS

abique:
I quickly went through and this is confusing.
It seems that you are mixing together two different goals:
- engine trigger event, like you may want a trigger for an envelope; this is engine side, audio-thread
- user interface actions, which may have an icon in a toolbar, be bound to shortcut, have a localized display name, a tooltip etc... and basically trigger code on the GUI (main thread)

To me it seems that you are mixing GUI and DSP.

include/clap/ext/draft/actions.h
```diff
@@ -0,0 +1,166 @@
...
+   // Flushes a set of action triggers.
+   // This method must not be called concurrently to clap_plugin->process().
+   //
+   // Note: if the plugin is processing, then the process() call will already achieve the
+   // triggers, so a call to flush isn't required, also be aware
+   // that the plugin may use the sample offset in process(), while this information would be
+   // lost within flush().
+   //
+   // [active ? audio-thread : main-thread]
+   void(CLAP_ABI *flush)(const clap_plugin_t        *plugin,
```
abique, line 103: trigger are stateless no need to flush

include/clap/ext/draft/actions.h
```diff
@@ -0,0 +1,166 @@
...
+   //
+   // The host will then schedule a call to either:
+   // - clap_plugin.process()
+   // - clap_plugin_actions.flush()
+   //
+   // This function is always safe to use and should not be called from an [audio-thread] as the
+   // plugin would already be within process() or flush().
+   //
+   // [thread-safe,!audio-thread]
+   void(CLAP_ABI *request_flush)(const clap_host_t *host);
```
abique, line 161: Not needed

Trinitou:
> I quickly went through and this is confusing. It seems that you are mixing together two different goals:
> 
> * engine trigger event, like you may want a trigger for an envelope; this is engine side, audio-thread
> * user interface actions, which may have an icon in a toolbar, be bound to shortcut, have a localized display name, a tooltip etc... and basically trigger code on the GUI (main thread)
> 
> To me it seems that you are mixing GUI and DSP.

I hope I could clear things up a bit. Now I went for the trigger event aspect first. For the other thing, I'll probably create a separate extension pull request.

Trinitou:
Also I took out the "enabled" flag for triggers as it should be added to parameters as well then. But this is a completely different topic. Started a new discussion (#234).

include/clap/events.h
```diff
@@ -113,6 +113,10 @@ enum {
    CLAP_EVENT_MIDI,       // raw midi event; clap_event_midi
    CLAP_EVENT_MIDI_SYSEX, // raw midi sysex event; clap_event_midi_sysex
    CLAP_EVENT_MIDI2,      // raw midi 2 event; clap_event_midi2
+
+   // Represents a trigger
+   // Uses clap_event_trigger.
+   CLAP_EVENT_TRIGGER,
```
abique, line 119: Use the same style as above for the description please

include/clap/ext/draft/triggers.h
```diff
@@ -0,0 +1,132 @@
...
+
+static CLAP_CONSTEXPR const char CLAP_EXT_TRIGGERS[] = "clap.triggers.draft/0";
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// @page Trigger events
+///
+/// This extension enables the plugin to provide a set of triggers to the host.
```
abique, line 14: Maybe expose instead of provide?

include/clap/ext/draft/triggers.h
```diff
@@ -0,0 +1,132 @@
...
+///
+/// Some examples for triggers:
+/// - start recording into a plugin-internal audio/note buffer
+/// - trigger a retrospective audio/note looper
+/// - trigger a sample-and-hold unit (maybe even per-voice)
+/// - retrigger an envelope during voice lifetime
```
abique, line 20: We don't need the "re" here, some envelopes may simply be triggered using the trigger event regardless of voice lifetime.

So I'd just say "trigger an envelope"

include/clap/ext/draft/triggers.h
```diff
@@ -0,0 +1,132 @@
...
+/// - shake event for a spring-reverb
```
abique, line 21: This example is interesting, because when you shake a spring reverb you may want an intensity, etc...

So two options, maybe this example doesn't fit or the trigger event needs additional data?

I think that the trigger is fine without data, and for triggering the spring, you may want instead to use a note port or notes on a specific channel.

include/clap/ext/draft/triggers.h
```diff
@@ -0,0 +1,132 @@
...
+   // When set:
+   // - trigger events can be recorded
+   // - trigger events can be played back
+   //
+   // The host can send live user trigger events regardless of this flag.
+   //
+   // If this trigger affects the internal processing structure of the plugin, ie: max delay, fft
+   // size, ... and the plugins needs to re-allocate its working buffers, then it should call
+   // host->request_restart(), and perform the change once the plugin is re-activated.
+   CLAP_TRIGGER_IS_AUTOMATABLE = 1 << 0,
```
abique, line 33: A trigger that's isn't automatable is pointless isn't it?

include/clap/ext/draft/triggers.h
```diff
@@ -0,0 +1,132 @@
...
+
+   // Does this trigger support per port automations?
+   CLAP_TRIGGER_IS_AUTOMATABLE_PER_PORT = 1 << 4,
+
+   // This trigger will affect the plugin output and requires to be done via
+   // process() if the plugin is active.
+   //
+   // A simple example would be a sample-and-hold, triggering it will change the output signal and must be
+   // processed.
+   CLAP_TRIGGER_REQUIRES_PROCESS = 1 << 5,
```
abique, line 52: Because the trigger is stateless, it isn't something that you can delay and send later on.
I think that the trigger must require process.
It is like a note event, you can't delay those, and they force processing.

include/clap/ext/draft/triggers.h
Trinitou: Also already thought about this point. What's your personal opinion? In the end we could achieve the trigger functionality with notes, just ignoring note-offs and all the data we don't need. So is a triggers extension useful/special enough to justify its existence?

abique:
Thank you for your changes, we're almost there. 👍

include/clap/events.h
Trinitou: What do you mean with 'style' in this case? Language or punctuation at the end of the line?

Trinitou:
Another question: Maybe 'trigger ports' would be an even better name for the extension?

include/clap/ext/draft/triggers.h
Trinitou: I think it's better to just remove that example to avoid confusion. So I'll do that.

include/clap/events.h
abique: Comment at the end of the line.
But actually forget what I said, as we use both styles. Just that I was just looking at the 3 line above in the github review thing, my bad.

include/clap/ext/draft/triggers.h
abique: Trigger is still cool, you can show a button for it in the host gui, ...

abique:
I think it is good as it is, I will merge the PR into the next branch and there might be more discussions about it later.

Thank you very much @Trinitou !

PR #238, opened 2022-12-07
Output CLAP version as STATUS

Schroedingers-Cat:
This changes the output of the CLAP version in CMake to a STATUS message. Before this change, CMake would print out the version as red text, making it look like an error.


baconpaul:
This change is correct - thanks - but I think the pr should go against next so we can stage it for 1.1.4 or so

Schroedingers-Cat:
> I think the pr should go against next so we can stage it for 1.1.4 or so

Ok, makes sense!

abique:
Thanks!

PR #241, opened 2022-12-12
clap 1.1.4 (draft)

abique:


include/clap/clap.h
```diff
@@ -56,8 +56,9 @@
 #include "ext/draft/file-reference.h"
 #include "ext/draft/midi-mappings.h"
 #include "ext/draft/preset-load.h"
-#include "ext/draft/quick-controls.h"
+#include "ext/draft/remote-controls.h"
 #include "ext/draft/state-context.h"
 #include "ext/draft/surround.h"
 #include "ext/draft/track-info.h"
 #include "ext/draft/tuning.h"
+#include "ext/draft/param-indication.h"
```
Bremmers, line 64: ext/draft/context-menu.h should be added as well I think?

include/clap/clap.h
```diff
@@ -56,8 +56,10 @@
 #include "ext/draft/file-reference.h"
 #include "ext/draft/midi-mappings.h"
 #include "ext/draft/preset-load.h"
-#include "ext/draft/quick-controls.h"
+#include "ext/draft/remote-controls.h"
 #include "ext/draft/state-context.h"
 #include "ext/draft/surround.h"
 #include "ext/draft/track-info.h"
```
Trinitou, line 62:
```suggestion
#include "ext/draft/track-info.h"
#include "ext/draft/triggers.h"
```
triggers extension needs to be added as well

include/clap/clap.h
```diff
@@ -56,8 +56,10 @@
...
 #include "ext/draft/tuning.h"
+#include "ext/draft/param-indication.h"
+#include "ext/draft/context-menu.h"
```
Trinitou, line 65: The rest of the file looks to me as if the includes are sorted alphabetically. Those two includes are out of order.

abique: I can sort them

include/clap/ext/draft/track-info.h
```diff
@@ -4,24 +4,48 @@
 #include "../../color.h"
 #include "../../string-sizes.h"
 
+// This extensions let the plugin query info about the track it's in.
+// It is useful when the plugin is created, to initialize some parameters (mix, dry, wet)
+// and pick a suitable configuartion regarding audio port type and channel count.
+
 static CLAP_CONSTEXPR const char CLAP_EXT_TRACK_INFO[] = "clap.track-info.draft/0";
```
Bremmers, line 11: Should be increased to .draft/1 now I think?

abique: Yes!

ChangeLog.md
```diff
@@ -1,3 +1,18 @@
+# Changes in 1.1.4
+
+* CMake: update some targets to link against `clap` instead of `clap-core`
+* [params.h](include/clap/ext/params.h): fix documentation typos, and missing items
+
+## Draft extensions
+
+* [audio-ports-activation.h](include/clap/ext/draft/audio-ports-activation.h): new draft extension which lets the host (de)activate audio ports
```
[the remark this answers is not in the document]
abique: @baconpaul do you have a suggestion?

ChangeLog.md
```diff
@@ -1,3 +1,18 @@
...
+* [context-menu.h](include/clap/ext/draft/context-menu.h): new draft extension which let the host and plugin exchange context menu entries and popup the menu
+* [param-indication.h](include/clap/ext/draft/param-indication.h): new draft extension which let the host inform the plugin that a parameter is currently mapped to a physical controller
+* [remote-controls.h](include/clap/ext/draft/remote-controls.h): new draft extension which replaces `quick-controls.h`
+* [resource-directory.h](include/clap/ext/draft/resource-directory.h): new draft extension which lets the plugin save resources in a directory provided by the host
```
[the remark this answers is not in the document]
abique: no

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+
+   // accelerator may be null
+   bool(CLAP_ABI *add_entry)(const struct clap_context_menu_builder *builder,
+                             const char                             *label,
+                             const clap_context_menu_accelerator_t  *accelerator,
+                             bool                                    is_enabled,
+                             clap_id                                 action_id);
+
+   // accelerator may be null
+   bool(CLAP_ABI *add_checkmark)(const struct clap_context_menu_builder *builder,
```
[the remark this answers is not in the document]
robbert-vdh: Following https://github.com/free-audio/clap/pull/221#discussion_r1045892651, `add_check_entry()` is probably both the least confusing and the most consistent option (every other context menu API I found calls these 'check items', so this would be consistent with our `add_entry()` for regular, non-checkable menu items).

abique:
I wonder if we should put the trigger event into a namespace until the extension is stable?

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+   bool(CLAP_ABI *perform)(const clap_host_t                *host,
+                           const clap_context_menu_target_t *target,
+                           clap_id action_id);
+
+   // [main-thread]
+   bool(CLAP_ABI *can_popup)(const clap_host_t *host);
+
+   // Shows the host popup menu for a given parameter.
+   // If the plugin is using embedded GUI, then x and y are relative to the plugin's window,
+   // otherwise they're absolute coordinate, and screen index might be set accordingly.
```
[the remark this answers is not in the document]
abique: We can solve that in the future by having the host providing a list of screen.

Bremmers:
Can't it be moved to trigger.h for the time being?

abique:
We'll do an event extension for now, also it is a good opportunity to test the feature! :)

include/clap/events.h
```diff
@@ -259,6 +263,20 @@ typedef struct clap_event_midi2 {
    uint32_t data[4];
 } clap_event_midi2_t;
 
+typedef struct clap_event_trigger {
+   clap_event_header_t header;
+
+   // target trigger
+   clap_id trigger_id; // @ref clap_trigger_info.id
+   void   *cookie;    // @ref clap_trigger_info.cookie
```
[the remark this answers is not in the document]
abique: Sometimes clang-format does a poor indentation :/ So I'm not so sure about doing it like this.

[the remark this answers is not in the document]
robbert-vdh: I think as long as everything's consistent it's all good. Also makes contributing much simpler. If you first need to reformat a whole bunch of unrelated code it makes reading diffs a lot more difficult. And then you can just set your text editor to format on save and everything will be kept nice and tidy.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+typedef struct clap_context_menu_accelerator_key {
+   uint32_t flags;
+   uint32_t key; // ASCII key
+} clap_context_menu_accelerator_key_t;
+
+// Describes an accelerator which can consist of a key strike combo
+typedef struct clap_context_menu_accelerator {
+   uint32_t combo_size;
+   clap_context_menu_accelerator_key_t keys[4];
+} clap_context_menu_accelerator_t;
```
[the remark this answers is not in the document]
abique: Yes this is just a display thing and nothing else.
Because the plugin will handle its windowing events itself, same for the host.

The benefit is that the way those accelerators are shown in the menu is consistent.

I'm OK to remove the accelerators, I think it clutters the interface.

include/clap/ext/draft/triggers.h
```diff
@@ -0,0 +1,111 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// @page Trigger events
+///
+/// This extension enables the plugin to expose a set of triggers to the host.
+///
+/// Some examples for triggers:
+/// - trigger an envelope which is independant of the notes
```
jaxter184, line 17: nit: "independ**a**nt" -> "independ**e**nt"

include/clap/ext/draft/context-menu.h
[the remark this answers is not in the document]
robbert-vdh: Yeah if it's just purely for display purposes I'd also be fine with removing it. We should have a proper vote for that first of course. Realistically the odds of, say, Surge XT's context menu showing relevant global Bitwig shortcut keys (especially since most context menus and context menu items will be specific to a parameter) are very slim. Or the other way around.

include/clap/ext/draft/context-menu.h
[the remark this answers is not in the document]
abique: `add_checked_entry()` then maybe?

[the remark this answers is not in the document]
robbert-vdh: I'd go with `add_check_entry()` since it may or may not be checked. Or checkbox heh, but other toolkits seem to strictly stick with the term check (through they do use the term radio button for radio button-style options).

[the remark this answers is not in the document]
defiantnerd: I agree with @robbert-vdh and @abique

include/clap/ext/draft/audio-ports-activation.h
```diff
@@ -0,0 +1,46 @@
...
+#endif
+
+typedef struct clap_plugin_audio_ports_activation {
+   // returns true if the plugin supports activation/deactivation while processing.
+   // [main-thread]
+   bool(CLAP_ABI *can_activate_while_processing)(const clap_plugin_t *plugin);
+
+   // activate the given port
+   // [main-thread]
+   uint32_t(CLAP_ABI *set_active)(const clap_plugin_t *plugin, bool is_input, uint32_t port_index, bool is_active);
```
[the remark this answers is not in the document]
abique: Then it needs to be called on both, and we'd need to introduce `set_active_on_audio_thread()`.

ChangeLog.md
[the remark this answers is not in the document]
baconpaul: A new extension which allows a host to inform a plugin whether an audio port is an active port in the host audio context, and allow the host to respond accordingly

something like that?

ChangeLog.md
```diff
@@ -1,3 +1,28 @@
...
+## Draft extensions
+
+### New
+
+* [audio-ports-activation.h](include/clap/ext/draft/audio-ports-activation.h): new draft extension which allows a host to inform a plugin whether an audio port is an active port in the host audio context, and allow the host to respond accordingly
+* [context-menu.h](include/clap/ext/draft/context-menu.h): new draft extension which let the host and plugin exchange context menu entries and popup the menu
+* [param-indication.h](include/clap/ext/draft/param-indication.h): new draft extension which let the host inform the plugin that a parameter is currently mapped to a physical controller
+* [remote-controls.h](include/clap/ext/draft/remote-controls.h): new draft extension which replaces `quick-controls.h`
+* [resource-directory.h](include/clap/ext/draft/resource-directory.h): new draft extension which lets the plugin save resources in a directory provided by the host
+* [triggers.h](include/clap/ext/draft/triggers.h): new draft extension which exposes triggers to the host, triggers are data less events
```
Trinitou, line 15:
```suggestion
* [triggers.h](include/clap/ext/draft/triggers.h): new draft extension which exposes triggers to the host, triggers are data-less events
```

include/clap/ext/draft/audio-ports-activation.h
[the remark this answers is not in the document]
tim-janik: For `can_activate_while_processing() == true`:

I need to second the concern here. In Anklang, *only* the audio-thread has authoritative knowledge over use/unuse of plugin ports. Calling set_active() from the main-thread means that the audio-thread (possibly having real-time scheduling prios) needs to queue the call into the main-thread (not having real-time scheduling prios) which might call set_active() when the use/unuse state in the audio-thread has become outdated/invalid.
Also, it is impossible to synchronize set_active() so it's not called concurrently while process() is being called in the audio-thread, which can make plugin impls considerably harder.

For `can_activate_while_processing() == false`, i.e. plugin must be "deactivated" for set_active():

I fail to see the use case here, at least in my DAW, plugins are loaded, activated and stay activated.
We cannot deactivate the plugin i.e. halt processing, just to set an optimization flag.

My recommendation:

I would have liked the ability to "flag" ports for being currently used/unused to be very tightly coupled to the process() function which runs in the audio-thread exclusively, that avoids thread synchronization issues and provides the flag exactly where/when it is used.
I.e. preferably introduce a new event type `CLAP_EVENT_PORT_FLAGS` so the host can pass this (and similar) infos into `process()` via the `in_events` queue, or alternatively introduce another function that is called right before process():

```c
#define CLAP_PORT_FLAG_UNUSED    0x01
  // configure the given port
  // [audio-thread]
   uint32_t (*set_port_flags) (const clap_plugin_t *plugin, bool is_input, uint32_t port_index, uint64 port_flags);
```

But IMHO, passing port flags via the `in_events` queue would be the simpler approach.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+   // accelerator may be null
+   bool(CLAP_ABI *add_checkmark)(const struct clap_context_menu_builder *builder,
+                                 const char                             *label,
+                                 const clap_context_menu_accelerator_t  *accelerator,
+                                 bool                                    is_enabled,
+                                 bool                                    is_checked,
+                                 clap_id                                 action_id);
+
+   bool(CLAP_ABI *add_separator)(const struct clap_context_menu_builder *builder);
+
```
tim-janik, line 51: Can this API be made more future proof (and simpler) by using flags? I.e.:

```c
typedef enum clap_context_menu_flags {
  CLAP_CONTEXT_MENU_ENTRY_DISABLED = 1,
  CLAP_CONTEXT_MENU_ENTRY_CHECKBOX = 2,
  CLAP_CONTEXT_MENU_ENTRY_CHECKED = 4,
  CLAP_CONTEXT_MENU_ENTRY_SEPARATOR = 8,
} clap_context_menu_flags_t;
bool (CLAP_ABI *add_entry) (const struct clap_context_menu_builder *builder,
                            const char                             *label,
                            clap_context_menu_flags_t               flags,
                            clap_id                                 action_id);
```

That allows adding _TITLE, _RADIO, and other menu item types in the future and uses 1 add-entry function instead of three.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+   bool(CLAP_ABI *populate)(const clap_plugin_t               *plugin,
+                            const clap_context_menu_target_t  *target,
+                            const clap_context_menu_builder_t *builder);
+
+   // If target is null, assume global context
+   // [main-thread]
+   bool(CLAP_ABI *perform)(const clap_plugin_t              *plugin,
+                           const clap_context_menu_target_t *target,
+                           clap_id                           action_id);
+} clap_plugin_context_menu_t;
```
tim-janik, line 75: To make sure I understand things right here, can clap_plugin_context_menu->populate() be called from the host on the plugin to query for menu items that the host can add to its plugin specific context menu that is shown in the host device panel? (without the plugin showing or having its own GUI)

And, I assume all of this API is intended to be [main-thread] only? It'd probably be good to state that somewhere.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,121 @@
...
+   // If the plugin is using embedded GUI, then x and y are relative to the plugin's window,
+   // otherwise they're absolute coordinate, and screen index might be set accordingly.
+   // If target is null, assume global context
+   // [main-thread]
+   bool(CLAP_ABI *popup)(const clap_host_t                *host,
+                         const clap_context_menu_target_t *target,
+                         int32_t                           screen_index,
+                         int32_t                           x,
+                         int32_t                           y);
+} clap_host_context_menu_t;
```
tim-janik, line 108: Unless I'm missing something, this API is meant for the plugin to request the host to show a context menu *on top* of its plugin GUI?

May I ask, when/why would a plugin trigger that?
Don't plugins that have their own GUI also have their own menus?

In case the plugin is meant to trigger a host menu popup in response to key presses or mouse button presses that the plugin received, the API lacks: a) the mouse button index (1,2,3); b) a time stamp.
Those two a crucial to correctly implement press-drag-release vs click-select popup menu behavior. Also the time stamp is crucial to sort out pointer capture logic (the X11 API has more details on this).

I'm not sure what exactly is needed on mac/win/wayland for this though, and I wonder if it's really worth opening this can of worms (mixing host menus with plugin GUIs), but then I'm assuming plugins with GUIs would have their own menus anyway...

include/clap/ext/draft/context-menu.h
jaxter184: My gut reaction is that this would impact performance, and that it gives extra runtime flexibility that most people won't need, but maybe that's OK? If I understand correctly, it only gets run a few times, and only ever in the main thread.

include/clap/ext/draft/context-menu.h
jaxter184: I was under the impression that it would be a context menu native to the host's UI, as pictured below. I'm not sure when the "global" menu would pop up, but I'm assuming the "param" menu will show up when you right-click a param knob in Bitwig's UI. I guess it's up to the host what they decide to do with the information, but as I understand it, this isn't intended to be used with a plugin's own UI. To that end, I'm not 100% sure what the purpose of the `popup` method in `clap_host_context_menu_t` is, but I'm guessing it has something to do with embedded UIs.
context menu

Trinitou: Already asked and answered here: https://github.com/free-audio/clap/pull/221#discussion_r1040583076

So maybe there should be explanatory comment in the code about what's the purpose/use case of the popu9 method?

include/clap/ext/draft/context-menu.h
abique: Your proposal as it is, doesn't work because it doesn't provide the checked state.
Different menu entries requires different data.

I think this menu builder is quite simple, but can probably do the job for the next 10 years.

include/clap/ext/draft/context-menu.h
abique: yes, and for the menu builder I think it is implied, but I can write it explicitly.

include/clap/ext/draft/context-menu.h
abique: popup as it names suggest is to be called by the plugin to ask the host to popup its menu for a given target.
The host can call plugin->populate() to fetch menu entries.

robbert-vdh: > Don't plugins that have their own GUI also have their own menus?

Most don't. The only VST3 plugin I know of that integrates the host's context menu into its own context menu is Surge XT. But there are many VST3 plugins that pop up the host's context menu when you right click on a parameter, because that's relatively simple to implement and adds a lot of utility to a plugin using a gesture that would otherwise be unused.

include/clap/ext/draft/audio-ports-activation.h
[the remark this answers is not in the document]
robbert-vdh: Yes, I concur that it's just not possible to synchronize this correctly if the main thread must communicate information that only the audio threads are guaranteed to have up to date information on.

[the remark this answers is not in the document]
abique: There is no need to synchronize correctly.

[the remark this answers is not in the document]
robbert-vdh: There is. The whole point of this extension is to allow plugins to skip processing auxiliary inputs and outputs when they're not connected. If those messages arrive early or late to the plugin then the plugin will produce different results compared to if it didn't implement the extension at all and unconditionally processed those inputs/outputs.

include/clap/ext/draft/audio-ports-activation.h
```diff
@@ -0,0 +1,50 @@
...
+///   compute it
+///
+/// Audio ports can only be activated or deactivated when the plugin is deactivated, unless
+/// can_activate_while_processing() returns true.
+///
+/// Audio buffers must still be provided if the audio port is deactivated.
+/// In such case, they shall be filled with 0 (or whatever is the neutral value in your context)
+/// and the constant_mask shall be set.
+///
+/// Audio ports are initially in the active state.
```
robbert-vdh, line 21: It may also be worth explicitlymentioning whether or not the audio port state persists being deactivated and reactivated. I was just wondering about this and it seems to me like it would persist since it nowhere says that it does not, but it probably doesn't hurt to be explicit about it and I can imagine hosts or plugins interpreting this differently.

include/clap/ext/draft/audio-ports-activation.h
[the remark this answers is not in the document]
abique: This state change is the consequence of a user action on the GUI.
This change doesn't need to be applied at the very same sample time on both sides.
If you want to have "reproduceable" results, then you should perform all your (de)activations before activating the plugin and then you render.

include/clap/ext/draft/audio-ports-activation.h
abique: I'm not sure if it should persist, but at least I'm sure that the host should explicitly set the state after loading the plugin, that'll avoid bugs.

include/clap/ext/draft/context-menu.h
tim-janik: > Your proposal as it is, doesn't work because it doesn't provide the checked state. Different menu entries requires different data.
> 
> I think this menu builder is quite simple, but can probably do the job for the next 10 years.

The checked state is flag 0x04.

include/clap/ext/draft/context-menu.h
tim-janik: > popup as it names suggest is to be called by the plugin to ask the host to popup its menu for a given target. The host can call plugin->populate() to fetch menu entries.

Can the host call `plugin->populate()` without a previous call to popup()?

That way the host can offer plugin specific actions in the host device panel menus without the plugin implementing or showing its own GUI.

include/clap/ext/draft/context-menu.h
abique: So what are the other entries that you'd like to add?

include/clap/ext/draft/audio-ports-activation.h
tim-janik: > I'm not sure if it should persist, but at least I'm sure that the host should explicitly set the state after loading the plugin, that'll avoid bugs.

What about compatibility here?
Say host-v1 doesn't support plugin port deactivation, it relies on plugins always having active ports.
Then host-v2 supports the extension, disables some ports and saves the state.
The project is shared with a system that runs host-v1. host-v1 loads the project/state, if the "deactivated" state persists, playback will be broken. So the state should definitely *not* persist.

So far I think the current draft is ill designed.
a) @abique may have a single use case in mind where port activation can be linked to a single user action and in a very limited scenario (plugin loading or deactivated engine) the port activation may be adjusted from the main thread.
b) However there are a number of cases (automatic routing, multi-channel support, noise gating, etc) where it'd really be useful to tell a plugin during process() on-the-fly if a specific port is currently in use or not. And that state is definitely temporary and for performance optimizations *only*.

The cases (b) can be all be solved with a new port activation event type, as can be case (a).
The current API proposal *only* supports case (a) though and is IMHO way more complex than a new event type.

include/clap/ext/draft/context-menu.h
Trinitou: The menu builder with methods looks a lot cleaner to me than flags. I think if one has to implement the methods, it doesn't leave so much  room for interpretation/bad combinations of flags. Flags would provide some data compression but only for a short moment during runtime. Most of the time one would probably directly unpack them again as bools.

But I also wonder a bit how a "radio" element would be added in the future. Maybe as a special kind of submenu where each entry must be a check entry which is then checked exclusively? Otherwise for a special radio element with an index value, perform would have to be called with an int value.

Trinitou (changes requested):

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,112 @@
...
+   // Adds a separator line.
+   bool(CLAP_ABI *add_separator)(const struct clap_context_menu_builder *builder);
+
+   // Starts a sub menu with the given label.
+   // Each begin_submenu() must have a corresponding end_submenu()
+   bool(CLAP_ABI *begin_submenu)(const struct clap_context_menu_builder *builder,
+                                 const char                             *label);
+
+   // Ends the current sub menu.
+   bool(CLAP_ABI *end_submenu)(const struct clap_context_menu_builder *builder);
```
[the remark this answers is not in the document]
Bremmers: I found the original code surprisingly easy to implement and consume. This proposal looks elegant (I was thinking in this direction myself too), but it actually complicates things for everyone.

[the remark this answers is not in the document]
Trinitou:
```suggestion
   // Starts a sub menu with the given label.
   bool(CLAP_ABI *begin_submenu)(const struct clap_context_menu_builder *builder,
                                 const char                             *label);

   // Ends the current sub menu.
   bool(CLAP_ABI *end_submenu)(const struct clap_context_menu_builder *builder);
```

Ok, this got me thinking further. The end_submenu method actually shouldn't be required then as the the caller of populate could be responsible for terminating "unending" sub menus nevertheless.

include/clap/ext/draft/context-menu.h
tim-janik: > So what are the other entries that you'd like to add?

Personally, I don't need any atm. ;-)

I've just designed and seen enough menu APIs in the past to have an idea what feature requests might come up over time. I've outlined some types above already, example:

image

This could be described by an array of:
```c
struct Entry {
  uint_t item_type; // TITLE, SEPARATOR, ITEM, RADIO, CHECKBOX
  clap_id_t id;           // identifier for activation
  uint64_t flags;       // DISABLED=0x1, CHECKED=0x2
  const char *const label; // unused for SEPARATOR
  const char *const icon; // unicode char, or "mi-<materialicon>", "fa-<fontawesomeicon>", etc
  const char *const hotkey; // just a hint, ignored by most hosts to avoid conflicts
  const char *const category; // either empty or the label of a menu item that is a submenu parent
  const char *const tooltip; // brief description (might need more than 256 chars in procedural API)
};
// Consecutive radio items form a group that has a single unique selection 
```

In a declarative API, the returned array and `char*` strings are assumed to be static const, i.e. persist while the plugin is loaded. The `category` label is a bit awkward for long lists of menu items, but it allows to keep all items in a flat declarative array. If a procedural API is chosen like yours, `begin_submenu() + end_submenu()` works fine, and of course the strings would be `char label[256];` instead of static const.
 
Don't expect all hosts to make use of the submenu facility though, e.g. Anklang will present items in a long vertically wrapped list that is horizontally scrollable and not support submenus any time soon. Handling large submenus  can easily get out of hand and be confusing for users, especially if created via generic APIs, rather than designed by a UI expert.
Note that a declarative API that returns a single array can make things much simpler (and faster) if long lists of menu items are used (e.g. for bank or preset selections). Especially if checked/disabled states are encoded with the rest of the menu entry information, because that essentially requires re-querying of the whole menu hierarchy for every popup, just to update the DISABLED and CHECKED booleans.

robbert-vdh: One of the original proposals had the plugin provide an array to the host. The downsides of doing it that way is that the plugin also needs to allocate storage for the array. With the builder style an intermediate data structure isn't needed. If the host represents the context menu as some kind of tree, the builder can immediately construct that tree.

Trinitou: > Don't expect all hosts to make use of the submenu facility though, e.g. Anklang will present items in a long vertically wrapped list that is horizontally scrollable and not support submenus any time soon. Handling large submenus can easily get out of hand and be confusing for users, especially if created via generic APIs, rather than designed by a UI expert.

I think the current extension proposal (with submenus) should work fine for everybody. I mean, everyone can interpret the submenus as desired. E.g. Anklang is totally free to put the submenus on the top level as well - maybe surrounded with separators and using the submenu labels as section titles.
Otherwise there could be a flag or method that tells if a menu builder supports submenus. If such a concept is established, maybe there also could be other flags in the future like "supports radio entries".

include/clap/ext/draft/context-menu.h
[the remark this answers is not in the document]
baconpaul: Why is end submenu not needed? How else would you know it ended?

include/clap/ext/draft/context-menu.h
abique: I've reworked the context menu to have a design that is more "future proof".

Regarding the icon, I like the option to use unicode chars, but otherwise I don't want to push requirements on the host to provide thousand of standard icons. I don't have enough knowledge with that to take a good decision.

hotkey, we decided to get rid of it to avoid confusion about if is was a way to bind a shortcut to the host.

abique: If we are happy with the design, I'd like then to move forward.

This extension will be published as draft, so we can still improve and add more data to the menu entries.

If someone has enough knowledge and wants to make a proposition, then please make a PR and consider what the host or plugin would have to do to ensure the availability of icons/unicode characters.

include/clap/ext/draft/context-menu.h
```diff
@@ -24,38 +24,68 @@ typedef struct clap_context_menu_target {
...
+// This object isn't thread-safe and must be used on the same thread as it was provided.
+typedef struct clap_context_menu_builder {
+   void *ctx;
+
+   // Adds an entry to the menu.
+   // entry_data type is determined by entry_kind.
+   bool(CLAP_ABI *add_item)(const struct clap_context_menu_builder *builder,
+                            clap_context_menu_item_kind_t           item_kind,
+                            const void                             *item_data);
 } clap_context_menu_builder_t;
```
Bremmers, line 93: Definitely more future proof.

Perhaps this change would make it easier to use, just like the current design. clap_context_menu_entry and clap_context_menu_submenu could be removed, and you wouldn't have to set up structs to add these item kinds.

   bool(CLAP_ABI *add_item)(const struct clap_context_menu_builder *builder,
                            clap_context_menu_item_kind_t           item_kind,
                            const char                             *label,
                            bool                                    is_enabled,
                            const void                             *extra_item_data);

include/clap/ext/draft/audio-ports-activation.h
abique: > ... So the state should definitely not persist.

Yes.

> b) ...

This is a good scenario but there are a bunch of implications here:
1. the plugin must be able to perform the activation / deactivation in a real-time context
1. DSP will have to communicate that state change back to the GUI
1. this optimization is difficult to implement both for the host and for the plugin, especially if it is an event which can happen at anytime (multiple time in a single process call); latency compensation also has to be considered. It could even be that in practice absolutely no one implements it.
1. it forces the plugin to pre-allocate all necessary buffers, while the current approach allows to only have what you need.
1. this design could also prevent some optimization, because it has a realtime semantic: "port (de)activation" happens at this given sample, while the current approach is more relaxed.

I don't buy this hardcore optimization arguments, and I doubt that any plugin vendor wants to deal with a port activation event which can happen in the middle of the process block.

While on the other hand, something that is more likely to be useful:
1. plugin gets the activation on the main thread
3. plugin needs to allocate some buffers and pre-compute some data
4. plugin passes this data in a lock free queue to its DSP

It is simpler because everything happens in the main thread all the time. In other plugin formats, ports activation only happens while the plugin is de-activated as far as I know, so it is already a huge improvement.

include/clap/ext/draft/context-menu.h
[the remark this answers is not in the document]
abique: I discard the whole suggestion, I think that the "stack" approach is easier.

include/clap/ext/draft/context-menu.h
abique: When you add a separator, label and enable arguments become irrelevant.

include/clap/ext/draft/audio-ports-activation.h
robbert-vdh: I don't agree with 1) and 3) being problems at all. The whole point of this extension is to make changes in the plugin's audio processing. So it must already be able to handle these changes in a realtime context. And the plugin must also already be able to pass information from the process function to the GUI in a realtime safe way if it wants to have a GUI. The audio thread is the only place where the information is known to be up to date. Unless you add a mutex around the process funciton that can be locked from the GUI, in which case it would no longer be realtime safe.

include/clap/ext/draft/context-menu.h
```diff
@@ -0,0 +1,146 @@
...
+
+typedef struct clap_context_menu_check_entry {
+   // text to be displayed
+   const char *label;
+
+   // if false, then the menu entry is greyed out and not clickable
+   bool        is_enabled;
+
+   // if true, then the menu entry will be displayed as checked
+   bool        id_checked;
```
Bremmers, line 67: should be is_checked

Bremmers (approved):

include/clap/ext/draft/audio-ports-activation.h
abique: Here a proposal:
- this extension remain as it is
- we somehow find a way to tell the plugin if an output buffer is being produced/consumed, that could be done by setting the audio buffer to nullptr which'd be equivalent to "high impedance" state, if the plugin says that it supports this behavior.

baconpaul: In the clap juce wrapper we are sort of assuming the buffers we ask for won't be null (I think surge checks anyway). So if we do this change of nullable buffers we would definitely want it opt in through the extension.

A "you are deactivated" main thread message which precedes the first "the buffer is now null" audio thread message makes sense to me though. But so does an event on or before the first block with a null buffer in the audio thread and audio has to tell gui.

but the one thing I surmise from this conversation is: if hosts are going to start using nullptr as a sentinel for deactivated buffers they should only do so with a host which has advertised the ability to handle that (through implementing this extension and advertising the appropriate handshake)

Hope that helps?

abique: Yes the handshake would be part of this extension, like we'd add `bool supports_null_audio_buffers()`.

abique: I'd have preferred to use a flag on the audio_buffer to indicate if it was computed or if it'll be consumed, but we don't have room for that.

baconpaul: so we have 32 bits for channel count. I wonder if we could use a high bit there. but that is very bug prone. Since the constant mask is 64 bits we sort of have an implicit no-more-than-64-channel part of the API.

but null is best sentinel with the api we have I agree.

abique: Actually it is problematic to have two threads which can toggle the same value, or we'd need to have some kind of `AND`...
@robbert-vdh you won!

PR #249, opened 2022-12-22
INCLUDE: clap/ext/draft/audio-ports-activation.h: disambiguate docs

tim-janik:
The old wording can be interpreted to mean is_active=true can be set on audio-thread if can_activate_while_processing=true, but still is_active=false is only possible on main-thread.

This patch fixes that, to clarify that is_active=true *and* is_active=false can be set on the audio-thread if can_activate_while_processing.

### DIFF
--- a/include/clap/ext/draft/audio-ports-activation.h
+++ b/include/clap/ext/draft/audio-ports-activation.h
@@ -7,9 +7,9 @@
 /// This extension provides a way for the host to activate and de-activate audio ports.
 /// Deactivating a port provides the following benefits:
 /// - the plugin knows ahead of time that a given input is not present and can choose
-///   an optimized computation path
+///   an optimized computation path,
 /// - the plugin knows that an output is not consumed by the host, and doesn't need to
-///   compute it
+///   compute it.
 ///
 /// Audio ports can only be activated or deactivated when the plugin is deactivated, unless
 /// can_activate_while_processing() returns true.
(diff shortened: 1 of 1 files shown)


include/clap/ext/draft/audio-ports-activation.h
```diff
@@ -39,10 +39,10 @@ typedef struct clap_plugin_audio_ports_activation {
 
    // Activate the given port.
    //
-   // It is only possible to activate on the audio-thread if can_activate_while_processing() returns
-   // true.
+   // It is only possible to activate and de-activate on the audio-thread if
+   // can_activate_while_processing() returns true.
    //
-   // [active ? audio-thread : main-thread]
+   // [can_activate_while_processing ? audio-thread : main-thread]
```
abique, line 45: This change is wrong.
If the plugin isn't active it has to be called on the main thread.
If the plugin is active either it can't be called or it has to be called on the audio thread.

include/clap/ext/draft/audio-ports-activation.h
```diff
@@ -39,10 +39,10 @@ typedef struct clap_plugin_audio_ports_activation {
 
    // Activate the given port.
    //
-   // It is only possible to activate on the audio-thread if can_activate_while_processing() returns
-   // true.
+   // It is only possible to activate and de-activate on the audio-thread if
```
abique, line 42: maybe we can write (de)activate to make it shorter?

abique (changes requested):

include/clap/ext/draft/audio-ports-activation.h
tim-janik: I see, I thought `active` was referring to the port `is_active` flag, but you meant it to refer to the `clap_plugin->activate()` state. For the record, `ext/audio-ports-config.h` uses the following wording:
```c
   // [main-thread,plugin-deactivated]
   bool(CLAP_ABI *select)(const clap_plugin_t *plugin, clap_id config_id);
```

include/clap/ext/draft/audio-ports-activation.h
tim-janik: I literally used the wording from line 7 in the same file.

PR #253, opened 2022-12-22
1.1.5

abique:


baconpaul:
I would love to consider #254 in this release also if possible (although it would need rebasing to merge).

src/plugin-template.c
```diff
@@ -104,9 +132,10 @@ static bool my_plug_init(const struct clap_plugin *plugin) {
    my_plug_t *plug = plugin->plugin_data;
 
    // Fetch host's extensions here
-   plug->hostLog = plug->host->get_extension(plug->host, CLAP_EXT_LOG);
-   plug->hostThreadCheck = plug->host->get_extension(plug->host, CLAP_EXT_THREAD_CHECK);
-   plug->hostLatency = plug->host->get_extension(plug->host, CLAP_EXT_LATENCY);
+   plug->host_log = plug->host->get_extension(plug->host, CLAP_EXT_LOG);
+   plug->host_thread_check = plug->host->get_extension(plug->host, CLAP_EXT_THREAD_CHECK);
+   plug->host_latency = plug->host->get_extension(plug->host, CLAP_EXT_LATENCY);
+   plug->host_state = plug->host->get_extension(plug->host, CLAP_EXT_STATE);
```
robbert-vdh, line 138: This template doesn't use them, but if this is meant as an idiomatic example then it's probably worthwhile to at least add a comment that you need to check whether the extension pointers are non-null before dereferencing them. Someone blindly copying this example may assume that the stored pointers are always valid (since most people never seem to check for null pointers unless explicitly told to).

Also, you may want to add explicit casts here. C allows implicitly casting between pointer types (`const void*` to `const clap_host_ext_t*`), but in C++ this is illegal and will result in a compiler error.

robbert-vdh (changes requested):
LGTM.

baconpaul:
Oh you merged right before I corrected the change log and rebased to single commit. That's OK but if you want we can squash all of my commits into a single one.

baconpaul:
All these changes look good to me though, and I'm happy for this to become 1.1.5, either with this transaction log or with a cleaned up / rebased one.

abique:
I don't have a problem with merge commit, and I actually like to have full history because, it happened many times that I had to understand why the code was written in such a way and looking at the history was the best approach to understand what devs had in mind and which approach they tried, how the code grew etc...

Bremmers (approved):

include/clap/host.h
```diff
@@ -15,9 +15,11 @@ typedef struct clap_host {
    const char *name;    // eg: "Bitwig Studio"
    const char *vendor;  // eg: "Bitwig GmbH"
    const char *url;     // eg: "https://bitwig.com"
-   const char *version; // eg: "4.3"
+   const char *version; // eg: "4.3", see plugin.h for advices on how to format the version
```
baconpaul, line 18: advice, not advices

baconpaul (approved):

baconpaul (approved):

include/clap/host.h
abique: Fixed

baconpaul (approved):

PR #254, opened 2022-12-22
Change VERSION macros to allow compares

baconpaul:
1. Add a CLAP_VERSION_x_DIGIT which is just 1, 2, 3
2. Make CLAP_VERSION_x be the uint cast of that
3. Add a CLAP_VERSION_NUMBER which is MAJOR << 16 + MINOR << 8 + REV so you can do a #if CLAP_VERSION_NUMBER <= 0x010105
4. Update CMakeLists to use DIGIT definitions


baconpaul:
OK I force pushed this rebased to next.

include/clap/version.h
```diff
@@ -20,11 +20,15 @@ typedef struct clap_version {
 }
 #endif
 
-#define CLAP_VERSION_MAJOR ((uint32_t)1)
-#define CLAP_VERSION_MINOR ((uint32_t)1)
-#define CLAP_VERSION_REVISION ((uint32_t)5)
+#define CLAP_VERSION_MAJOR_DIGIT 1
```
abique, line 23: Maybe it should be digits, it could go beyond 9.

abique (changes requested):
Please add changelog entry as well.
This change will break all build systems which are currently parsing version.h (we do it in bitwig), so it has to be the very last time we change it.
Did you consider the option to simply remove the the cast to `uint32_t`?

include/clap/version.h
```diff
@@ -20,11 +20,15 @@ typedef struct clap_version {
...
+#define CLAP_VERSION_MINOR_DIGIT 1
+#define CLAP_VERSION_REVISION_DIGIT 5
+#define CLAP_VERSION_MAJOR ((uint32_t)CLAP_VERSION_MAJOR_DIGIT)
+#define CLAP_VERSION_MINOR ((uint32_t)CLAP_VERSION_MINOR_DIGIT)
+#define CLAP_VERSION_REVISION ((uint32_t)CLAP_VERSION_REVISION_DIGIT)
 #define CLAP_VERSION_INIT                                                                          \
    { CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION }
+#define CLAP_VERSION_NUMBER (CLAP_VERSION_MAJOR_DIGIT << 16) + (CLAP_VERSION_MINOR_DIGIT << 8) + CLAP_VERSION_REVISION_DIGIT
```
abique, line 31: You need `()` around the whole expression here

abique:
Given how CLAP_VERSION_NUMBER is constructed, I think if we use a C++ compiler we could `static_assert` that both the minor and revision are less than 256.

include/clap/version.h
abique: Actually you may also want to force the type to `uint32_t`?

abique: You could also use `CLAP_VERSION_MAJOR` in the construction instead of the digit one.

abique: One other idea would be to have:
```C
static int clap_version_cmp(const clap_version_t *v1, const clap_version_t *v2);
```

baconpaul: The problem is we want a compile time check so a function won’t do it. That’s why I had to knock out the uints since they confused the pre processor. But yea I definitely need parens 

the other option which may be way less disruptive is just define a clap-version-number as 0x010105 by hand and add a failure to the compile test in c++17 with a static assert. Then we don’t have to change folks who parse this 

if folks parse it we should add that as a comment too :)

lemme know which one you prefer and I can change it tomorrow am nyc time

robbert-vdh: A constexpr function would work for C++. Not for C of course

baconpaul:
> Given how `CLAP_VERSION_NUMBER` is constructed, I think if we use a C++ compiler we could `static_assert` that both the minor and revision are less than 256.

Yea very good idea 

the whole problem here stems from the explicit uint32 in the version ifdef which breaks pre processor math. That’s why I separated into digits and so on - but adding a bunch of static asserts to the test program is a great idea that I’ll do once you let me know if you want to do this or just maintain the version twice

include/clap/version.h
baconpaul: Can you skip a `#include` with a constexpr function? That was the original motivation here.

baconpaul:
Anyway I pushed a new version which responds to the review comments if we want to go this route. If we want to go another route it obviously won't :)

The far far easier version of this change is to just maintain a separate CLAP_VERSION_NUMBER as 0x010105L by hand in the header file and then use the static assert to fail to compile if you don't keep them in sync. If you prefer that version let me know and I'll submit a separate PR for that.

include/clap/version.h
abique: Why do you need to skip a `#include` based upon clap version?

include/clap/version.h
```diff
@@ -20,16 +20,34 @@ typedef struct clap_version {
...
 clap_version_is_compatible(const clap_version_t v) {
    // versions 0.x.y were used during development stage and aren't compatible
    return v.major >= 1;
 }
+
+
+#if defined(__cplusplus) && __cplusplus >= 201703L
+// Static assert some version constraints
+static_assert(CLAP_VERSION_MAJOR_DIGITS < 256 && CLAP_VERSION_MAJOR_DIGITS >= 1);
+static_assert(CLAP_VERSION_MINOR_DIGITS < 256 && CLAP_VERSION_REVISION >= 1);
```
abique, line 50: minor can be 0

abique:
If I understand well, the problem that you want to solve is:

```C
#if CLAP_VERSION_GE(1, 1, 4)
# include <x.h>
#endif
```

abique:
```C
#define A 5
#define B 6
#define C 8

#define CHECK_GE(X, Y, Z) ((X) > (A)) || ((X) == (A) && ((Y) > (B) || ((Y) == (B) && ((Z) >= (C)))))

#if CHECK_GE(7, 9, 9)
# error "Working!"
#else
# error "Not working :'("
#endif
```

```bash
clang -c tutu.c 
tutu.c:8:3: error: "Working!"
# error "Working!"
  ^
1 error generated.
```

abique:
So what I'd suggest:
- remove the `uint32_t` cast, so we have instead: `#define CLAP_VERSION_MAJOR 1`
- define a macro `CLAP_VERSION_GE(X, Y, Z)`
- we don't need `CLAP_VERSION_NUMBER`

I think we solve all the macro issues with that. If we need further utilities we should be able to write them using static inline / constexpr functions.

What do you think?

robbert-vdh:
What was the _actual_ reason for wanting to check the CLAP version using the preprocessor @Arakula? I feel like this may be an X-Y problem. If you just want to check if an include exists, you can use the standard [`has_include(<header/name.h>)`](https://en.cppreference.com/w/cpp/preprocessor/include) expression. And you can't keep all CLAP draft extensions around without symbol clashes anyways since draft 1, draft 2, draft 3, and the final version will likely use the same symbol names. So you'd need to vendor and modify those headers anyways.

include/clap/version.h
baconpaul: That was the original discussion which spawned this. Trying to make code which works with draft extensions moving around 

https://github.com/free-audio/clap/discussions/252#discussioncomment-4478704

baconpaul:
> So what I'd suggest:
> 
> * remove the `uint32_t` cast, so we have instead: `#define CLAP_VERSION_MAJOR 1`
> * define a macro `CLAP_VERSION_GE(X, Y, Z)`
> * we don't need `CLAP_VERSION_NUMBER`
> 
> I think we solve all the macro issues with that. If we need further utilities we should be able to write them using static inline / constexpr functions.
> 
> What do you think?

Isn’t the uint32 cast needed so it compiles properly in msvc when used in an initializer list?

obviiusky if that cast isn’t needed there’s way easier solutions. But I think it is for the clap version macros no?

also the solution here of a single long you compare with is what c++ compiler version, juce and vst3 all use today. But we can do something else sure!

baconpaul:
> What was the _actual_ reason for wanting to check the CLAP version using the preprocessor @Arakula? I feel like this may be an X-Y problem. If you just want to check if an include exists, you can use the standard [`has_include(<header/name.h>)`](https://en.cppreference.com/w/cpp/preprocessor/include) expression. And you can't keep all CLAP draft extensions around without symbol clashes anyways since draft 1, draft 2, draft 3, and the final version will likely use the same symbol names. So you'd need to vendor and modify those headers anyways.

If your claim is “there is never a reason to compile time check clap version” then (1) yes absolutely we don’t need this and (2) whoooeee I would ask you to show your work a bit more :)

robbert-vdh:
I'm not saying that. At all. I'm saying that in the case of loading old draft extensions there's probably a better way to do this especially when considering symbol clashes.

abique:
> Isn’t the uint32 cast needed so it compiles properly in msvc when used in an initializer list?
>
> obviiusky if that cast isn’t needed there’s way easier solutions. But I think it is for the clap version macros no?

We can put the cast into `CLAP_VERSION_INIT` also. If that doesn't work the github tests will catch it I suppose.
 
> also the solution here of a single long you compare with is what c++ compiler version, juce and vst3 all use today. But we can do something else sure!

We already have the major, minor and revision defines, so if we add the number, it means that it has to be constructed by using major, minor and revision defines. For me it is important to have a single definition of the version and it has to be easy to parse using a regex. Extracting the version from `0x01010c` seems more difficult to me, also it is ambiguous if hex should be used or if `0x010112` for `minor=12`. It could also be written as `1001012`: `Major * 1000000 + Minor * 1000 + Revision`.

I like `CLAP_VERSION_GE(1, 1, 4)` because it doesn't require to learn how the version is encoded, the code speaks for itself. `CLAP_VERSION_NUMBER >= 0x010104` is OK too.

> also the solution here of a single long you compare with is what c++ compiler version, juce and vst3 all use today.

Yes the format `0x0000000` has been used widely, but mac also used `__builtin_available(macOS 11.0, *)` which looks a bit more modern style to me.

Arakula:
> What was the _actual_ reason for wanting to check the CLAP version using the preprocessor @Arakula? I feel like this may be an X-Y problem. If you just want to check if an include exists, you can use the standard [`has_include(<header/name.h>)`](https://en.cppreference.com/w/cpp/preprocessor/include) expression. And you can't keep all CLAP draft extensions around without symbol clashes anyways since draft 1, draft 2, draft 3, and the final version will likely use the same symbol names. So you'd need to vendor and modify those headers anyways.

I tried to send an answer per email, but somehow this doesn't seem to make its way to Github (at least not to this discussion), so here it is again (sorry if it appears twice in the end):

The actual reason can be found here:
https://github.com/free-audio/clap/discussions/252

... and I'm afraid has_include() won't be enough.

For one, it's a C++17 thing, and I regularly work with older (and not necessarily C++) compilers. But that's just me ...

The other, more general reason is that V1.1.4 not only does away with 2 draft extensions, it also modifies one (track_info.h) - and to keep both versions supported means that the old header contents have to be kept in a modified form.

baconpaul:
> > Isn’t the uint32 cast needed so it compiles properly in msvc when used in an initializer list?
> > obviiusky if that cast isn’t needed there’s way easier solutions. But I think it is for the clap version macros no?
> 
> We can put the cast into `CLAP_VERSION_INIT` also. If that doesn't work the github tests will catch it I suppose.
> 
> > also the solution here of a single long you compare with is what c++ compiler version, juce and vst3 all use today. But we can do something else sure!
> 
> We already have the major, minor and revision defines, so if we add the number, it means that it has to be constructed by using major, minor and revision defines. For me it is important to have a single definition of the version and it has to be easy to parse using a regex. Extracting the version from `0x01010c` seems more difficult to me, also it is ambiguous if hex should be used or if `0x010112` for `minor=12`. It could also be written as `1001012`: `Major * 1000000 + Minor * 1000 + Revision`.
> 
> I like `CLAP_VERSION_GE(1, 1, 4)` because it doesn't require to learn how the version is encoded, the code speaks for itself. `CLAP_VERSION_NUMBER >= 0x010104` is OK too.
> 
> > also the solution here of a single long you compare with is what c++ compiler version, juce and vst3 all use today.
> 
> Yes the format `0x0000000` has been used widely, but mac also used `__builtin_available(macOS 11.0, *)` which looks a bit more modern style to me.

So the pr I sent does constructs the number from major minor and revision right?

thenthing about ge is you need to define a le gt lt EQ ne version too which is why I think most systems define a hex constant the way the pr did?  

thst built in available compiler intrinsic is an oddity of gcc and clang by the way (and actually caused surge for rack to fail to load on mac when using the initial rack cross compile environment even though it built). 

But I’m happy with whatever you think is best! I just tossed in what I thought was a simple pr whcih didn’t change the cast on current ifdefs (so won’t break anything except regex parsers) and let us define a constant we can compare with.  What do you think we should do?

baconpaul:
I also pushed a change which has the minor version >= 0 not 1. Thanks!

tim-janik:
I may be missing something here, but...
Can't we simply do what is best practice in other C projects?

I.e. (adapting from Gtk+ here):

```c
#define CLAP_VERSION_MAJOR    1
#define CLAP_VERSION_MINOR    1
#define CLAP_VERSION_REVISION 5
#define CLAP_CHECK_VERSION(major,minor,revision)    \
    (CLAP_VERSION_MAJOR > (major) || \
     (CLAP_VERSION_MAJOR == (major) && CLAP_VERSION_MINOR > (minor)) || \
     (CLAP_VERSION_MAJOR == (major) && CLAP_VERSION_MINOR == (minor) && \
      CLAP_VERSION_REVISION >= (revision)))
```

If a specific type is needed in some context, simply do:

```c
auto version = { (uint) CLAP_VERSION_MAJOR, (uint) CLAP_VERSION_MINOR, (uint) CLAP_VERSION_REVISION };
```

And if some project relies on being able to check < 1.1.5 with the pre-processor (since we've already released <=1.1.4 version macros with the broken type cast), they can at least do:
```c
#ifndef CLAP_CHECK_VERSION    // clap < 1.1.5
```

baconpaul:
Yea there’s two patterns. The macro or the compiled constant. We should pick one.  The macro is harder to do range comparisons and stuff (if version < 7 and > 608 is one we use in juce for instance)

And then decide if we want to strip the uint cast and potentially break people who rely on it in c or not

my choice was
- single constant like c++ juce and vst3
- Don’t strip the uint 

Which is what the pr here does. 

But happy to close this and someone to toss in an alternative

robbert-vdh:
> I tried to send an answer per email, but somehow this doesn't seem to make its way to Github (at least not to this discussion), so here it is again (sorry if it appears twice in the end):
> 
> The actual reason can be found here: #252
> 
> ... and I'm afraid has_include() won't be enough.
> 
> For one, it's a C++17 thing, and I regularly work with older (and not necessarily C++) compilers. But that's just me ...
> 
> The other, more general reason is that V1.1.4 not only does away with 2 draft extensions, it also modifies one (track_info.h) - and to keep both versions supported means that the old header contents have to be kept in a modified form.

Isn't it much more maintainable to then _always_ vendor the extensions? Instead of including either a vendored copy or the version from the CLAP headers depending on which CLAP version you're targetting?

<hr>

And wrt version comparisons, I'd just define a couple comparison macros. Then you don't need to encode the CLAP version as a single integer, and the usage is much more readable once you go beyond to 0-9 range for major/minor/revision version numbers. A `CLAP_VERSION_GE(x, y, z)`,  `CLAP_VERSION_LT(x, y, z)`, and a `CLAP_VERSION_EQ(x, y, z)` should cover most use cases, no?

baconpaul:
Ok and what would you do with the uint casts?

robbert-vdh:
I'd move them to the `CLAP_VERSION_INIT` macro.

baconpaul:
> I'd move them to the `CLAP_VERSION_INIT` macro.

OK! Should I mark that as a potentially breaking change in the changelog?

robbert-vdh:
Yeah I guess it should be mentioned on the changelog since with that change the `CLAP_VERSION_{MAJOR,MINOR,REVISION}` macros are now plain integer literals instead of `uint32_t` expressions. Though in practice of course the odds of this breaking anything are very low of course.

baconpaul:
OK well if you could re-review the change I think it now matches the macro style properly, has an appropriate change log, and tests the macros in the test builds.

baconpaul:
by the way I notice the checks don't run on a pull request. Is that on purpose? I can toss in the yml to fix that on a separate PR if you want.

src/main.cc
```diff
@@ -2,8 +2,24 @@
 
 // The purpose of this file is to check that all headers compile
 
+#if CLAP_VERSION_LT(CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION)
+SOFTWARE ERROR: CLAP_VERSION_LT is inconsistent
```
robbert-vdh, line 6: You can use the `#error` pragma for this. That prints the erorr message instead of resulting in a probably less readable parsing error.

robbert-vdh (changes requested):

src/main.cc
baconpaul: I thought some compilers didn’t support the error pragma - especially older msvc - hence the choice. But happy to change it sure!

baconpaul: Oh but in this spot it doesn't matter anyway since its a c++ file we just compile as a test. (I originally had these in the header then moved them). Anyway pushed change.

robbert-vdh: Some older compilers indeed may not support it, but then you'd just get an unsupported pragma warning or error with a reference to the line.  That's probably still better than the compiler just chugging along and trying to make sense of the file with those error messages as plain text. Microsoft's documentation doesn't go back before Visual Studio 2015, but that version does have `#error`. And if one compiler doesn't support it, it's probably MSVC.

Arakula: #error is supported back until Visual Studio 98, at least ... I'd have to fire up a virtual XP to check whether it works in Visual C++ 1.52 as well, but am pretty sureit does. I seem to remember that it already worked in Microsoft C 1.00 which I had in 1983 (I got no possibility to verify that anymore, though). The whole #pragma stuff is something that came in later (and where Microsoft and GNU C/C++ have severe syntactical disagreements).

baconpaul: Yeah but also in this context - where I added the error to the test case not the header file - our compiler set is smaller so they *definitely* support it.

Anyway curious how folks feel about the version as submitted here now. I think it meets all review constraints yeah?

robbert-vdh (approved):
LGTM.

include/clap/version.h
```diff
@@ -20,11 +20,18 @@ typedef struct clap_version {
...
+
 #define CLAP_VERSION_INIT                                                                          \
-   { CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION }
+   { (uint32_t)CLAP_VERSION_MAJOR, (uint32_t)CLAP_VERSION_MINOR, (uint32_t)CLAP_VERSION_REVISION }
+
+#define CLAP_VERSION_LT(maj,min,rev) ((maj < CLAP_VERSION_MAJOR) || \
+                    (maj == CLAP_VERSION_MAJOR && min < CLAP_VERSION_MINOR ) || \
+                    (maj == CLAP_VERSION_MAJOR && min == CLAP_VERSION_MINOR && rev < CLAP_VERSION_REVISION))
+#define CLAP_VERSION_EQ(maj,min,rev) ((maj == CLAP_VERSION_MAJOR) && (min == CLAP_VERSION_MINOR) && (rev == CLAP_VERSION_REVISION))
+#define CLAP_VERSION_GE(maj,min,rev) (!CLAP_VERSION_LT(maj,min,rev))
```
abique, line 34: Do we really need LT and EQ?

abique: `LT=!GE` right?

robbert-vdh: I think they're useful to have, and having them around doesn't really cost you anything.  `CLAP_VERSION_LT(x, y z)` has mcuh clearer semantics than writing out `!CLAP_VERSION_GE(x, y, z)` yourself. `EQ` is really specific of course but I can see it being useful if we ever ship a single borked CLAP revision.

baconpaul: With JUCE I've used all the operators including range comparisons. Which is why a number is nice but if we are going with macros, yeah, with < and == you can get everything; and GE is there just since it is a common case.

but we can do whatever you want. We can also provide no macros if you prefer and users can just write them themselves if they need, once the uint is stripped from the version. Up to you.

abique: In your macro, you need brackets around `(maj) < CLAP_VERSION_MAJOR`

baconpaul: > `LT=!GE` right?

I mean, the code 4 lines above your comemnt shows indeed, that GE == !LT, since that's how I coded it :)

src/main.cc
```diff
@@ -2,8 +2,24 @@
 
 // The purpose of this file is to check that all headers compile
 
+#if CLAP_VERSION_LT(CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION)
+#error CLAP_VERSION_LT is inconsistent
+#endif
+
+#if !CLAP_VERSION_EQ(CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION)
+#error CLAP_VERSION_EQ is inconsistent
```
abique, line 10: If the intention is to cover the _EQ with "unit tests", then you may also want to check:
- `CLAP_VERSION_EQ(CLAP_VERSION_MAJOR + 1, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION)`
- `CLAP_VERSION_EQ(CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR + 1, CLAP_VERSION_REVISION)`
- - `CLAP_VERSION_EQ(CLAP_VERSION_MAJOR, CLAP_VERSION_MINOR, CLAP_VERSION_REVISION + 1)`

abique:
> by the way I notice the checks don't run on a pull request. Is that on purpose? I can toss in the yml to fix that on a separate PR if you want. [edit: oh but now i see the check has an odd vcpkg thing it does so i don't know if that will work. would be great though if the PR could run the compiles for me!]

This is bad, I'll look into why.

src/main.cc
baconpaul: Oh that's smart. 

I just pushed the paren change. I will push that now.

include/clap/version.h
abique: > I mean, the code 4 lines above your comemnt shows indeed, that GE == !LT, since that's how I coded it :)

Yeah sorry, my comment was poorly formulated, I wanted to say that if we have GE we don't need anything else I suppose, but I've removed that comment, unfortunately you were too fast ;-)

baconpaul:
> > by the way I notice the checks don't run on a pull request. Is that on purpose? I can toss in the yml to fix that on a separate PR if you want. [edit: oh but now i see the check has an odd vcpkg thing it does so i don't know if that will work. would be great though if the PR could run the compiles for me!]
> 
> This is bad, I'll look into why.

Oh the reason why is because the trigger is [push, workflow_request], not [push, pull_request, workflow_request] in the .yml

src/main.cc
baconpaul: OK jut pushed.

include/clap/version.h
baconpaul: Pushed the bracket change.

abique:
> Oh the reason why is because the trigger is [push, workflow_request], not [push, pull_request, workflow_request] in the .yml

I've added it

baconpaul:
> > Oh the reason why is because the trigger is [push, workflow_request], not [push, pull_request, workflow_request] in the .yml
> 
> I've added it

OK let me rebase this then and we can see if they run.

ChangeLog.md
```diff
@@ -8,6 +8,10 @@
 * Remove UTF-8 BOM from a few files
 * [plugin-template.c](src/plugin-template.c): add state impl and some comments
 * [audio-ports-activation.h](include/clap/ext/draft/audio-ports-activation.h): improved documentation
+* [version.h](include/clap/version.h): Add a CLAP_VERSION_GE(maj,min,rev), _EQ and _LT macro. Remove the
```
abique, line 11: Maybe use the back quotes for the programming names?

baconpaul:
Great and with the rebase you can see the builds running! Super.

abique (approved):

abique:
Thanks!

ChangeLog.md
baconpaul: Sure just did that and squashed down all my commits and force pushed.

PR #263, opened 2022-12-26
Document more precise use of param name vs module

SamWindell:
Following the discussion: https://github.com/free-audio/clap/discussions/248#discussion-4683110

There was a slight consenus towards ensuring that the `name` field does not also contain the text of the `module`. Instead, the host should somehow display `module` alongside `name` if `name` alone would be too vague.

This PR suggests putting these rules into the header comments. It's separate from my other PR as it might be more controversial.

### DIFF
--- a/include/clap/ext/params.h
+++ b/include/clap/ext/params.h
@@ -221,11 +221,13 @@ typedef struct clap_param_info {
    //    per event.
    void *cookie;
 
-   // the display name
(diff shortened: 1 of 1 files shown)


abique:
Please add changelog entries.

abique:
> Please add changelog entries.

Never mind, I'll do it.

abique:
Thank you for both pull requests 👍

PR #270, opened 2023-01-02
1.1.7

abique:


abique:
One open question was if `preset-discovery.h` should be in `ext/` or not.

robbert-vdh:
> One open question was if `preset-discovery.h` should be in `ext/` or not.

It's not a plugin extension, but maybe there's another good way to organize this and other factories like the invalidation factory, converters, and whatever else is needed in the future? With the current setup a single subdirectory for all factories would suffice. Then you could even have a draft directory in there to separate draft factories from non-draft factories.

Trinitou:
> > > One open question was if `preset-discovery.h` should be in `ext/` or not.
> > 
> > 
> > It's not a plugin extension, but maybe there's another good way to organize this and other factories like the invalidation factory, converters, and whatever else is needed in the future? With the current setup a single subdirectory for all factories would suffice. Then you could even have a draft directory in there to separate draft factories from non-draft factories.
> 
> right so i like the idea of directories for things. `ext/` for extensions `factory/` for factories. And since this is a factory perhaps it belongs there?

Maybe like this?

- factory/(draft/)
  - plugin-instantiation (instead of plugin-factory)
  - plugin-invalidation
  - preset-discovery

abique:
maybe we need to make those plural?
- `factories/`
- `exts`

robbert-vdh:
I'd stick with singular. Using singular for namespaces and directories seems fairly common in most languages, with C# being the only language that comes to mind where the plural form is preferred. But ultimately it's just an aesthetic choice, so it doesn't really matter that much.

robbert-vdh:
This comment probably also needs to be updated: https://github.com/free-audio/clap/blob/18c4c37f195705e4246d0bf2128abce87e6099a3/include/clap/entry.h#L56

A nice thing about moving the plugin factory to a `factory` directory is that it's puts more emphasis on the fact that `entry.get_factory()` can be used for more than just the plugin factory. Since some people were unconditionally returning their plugin factory from that function.

baconpaul:
> 

I think that is smart but moving the base factory mag be a 1.2 thing perhaps? Also agree with Robbert that singular is fine and often used

abique:
> I think that is smart but moving the base factory mag be a 1.2 thing perhaps? Also agree with Robbert that singular is fine and often used

I think it is fine, it doesn't break the ABI, and I suspect that all of the projects simply includes `clap.h`.

baconpaul:
> > I think that is smart but moving the base factory mag be a 1.2 thing perhaps? Also agree with Robbert that singular is fine and often used
> 
> I think it is fine, it doesn't break the ABI, and I suspect that all of the projects simply includes `clap.h`.

Cool! I agree that as long as we mark the renames in the changelog should be fine

include/clap/factory/draft/plugin-invalidation.h
```diff
@@ -1,7 +1,12 @@
...
 
-#include "private/std.h"
-#include "private/macros.h"
+#include "../../private/std.h"
+#include "../../private/macros.h"
+
+// Use it to retrieve const clap_plugin_invalidation_factory_t* from
+// clap_plugin_entry.get_factory()
+static const CLAP_CONSTEXPR char CLAP_PLUGIN_INVALIDATION_FACTORY_ID[] =
+   "clap.plugin-invalidation-factory/draft0";
```
Trinitou, line 9:
```suggestion
   "clap.plugin-invalidation-factory/draft/0";
```
use same naming convention as draft extensions

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,254 @@
...
+
+#include "../../private/std.h"
+#include "../../private/macros.h"
+#include "../../string-sizes.h"
+#include "../../version.h"
+
+// Use it to retrieve const clap_preset_discovery_factory_t* from
+// clap_plugin_entry.get_factory()
+static const CLAP_CONSTEXPR char CLAP_PRESET_DISCOVERY_FACTORY_ID[] =
+   "clap.preset-discovery-factory/draft-0";
```
Trinitou, line 40:
```suggestion
   "clap.preset-discovery-factory/draft/0";
```

ChangeLog.md
```diff
@@ -1,3 +1,12 @@
+# Changes in 1.1.7
+
+* Add a [factory](incl/../include/clap/factory) folder for better organization.
+
+## Draft extensions
+
+* [preset-discovery.h](include/clap/preset-discovery.h): new extension which allows the host to index the plugin's preset which are stored on disk.
+* [context-menu.h](include/clap/ext/dra/../draft/context-menu.h): add "title" menu entry
```
Trinitou, line 8:
```suggestion
* [context-menu.h](include/clap/ext/dra/../draft/context-menu.h): add "title" menu entry

## Draft factories

* [preset-discovery.h](include/clap/factory/draft/preset-discovery.h): new factory which allows the host to index the plugin presets which are stored on disk.
```

Trinitou (changes requested):
For more consistency between factory and extension type names, I'd suggest the following changes to the factory type names:

- remove the 'factory' part., e.g. plugin_invalidation_factory_t -> plugin_invalidiation_t. (the extension type names also don't contain an explicit 'extension' part)
- in the same manner plugin_factory_t then should then become plugin_instantiation_t (or plugin_creation_t, plugin_discovery_t or similar). (although I'm not sure whether that might be a breaking change)

ChangeLog.md
Trinitou: A bit off-topic: How long are old changelog hyperlinks supposed to work? Or should they rather be maintained or removed with newer versions or whenever mentioned files are moved/renamed?
In general the '..' notation might reduce the amount of maintenance work or at least extend the lifetime of the hyperlinks.

abique:
I'm happy with `clap_preset_discovery_factory_t` as a name.

ChangeLog.md
abique: They should work with the version at which they were released.
We won't update the past changelog entries after a release.

abique:
There are multiple inconsistencies in the naming of the draft identifiers over the project, but we should not fix them now because it'd create a break-point.

I don't think it is a very important thing, and at the end of the day if the extension becomes stable then the draft disappears.

abique:
I wondering if with preset discovery, the flags should be set for the preset "as well or instead" of the location?

include/clap/ext/draft/preset-load.h
```diff
@@ -9,19 +9,13 @@ extern "C" {
...
-   // [main-thread]
-   bool(CLAP_ABI *from_file)(const clap_plugin_t *plugin, const char *path);
-
-   // Loads a preset in the plugin native preset file format from a path which points to a preset
-   // container file.
-   // preset_id must be a valid string, which is specific to the plugin itself and identifies the
-   // desired preset within the given preset container.
+   // Loads a preset in the plugin native preset file format from a URI. eg:
+   // - "file:///home/abique/.u-he/Diva/Presets/Diva/HS Bass Nine.h2p"
+   // - "file:///home/abique/my-sound-bank/<preset-id>"
```
robbert-vdh, line 14: How about using a colon for targetting presets inside of bank files? Kinda like how you specify a path on a remote machine (`user@hostname:path/to/file`). Would that remove a layer of ambiguity (you now need to first check if `/home/abique/my-sound-bank` is a file or directory) or would that only make things more complex? Then the example would become `file:///home/abique/my-sound-bank:<preset-id>`.

abique: I think it becomes then more difficult to parse isn't it?
Because `:` is a valid filename character on Linux while `/` isn't.
I believe that using `/` makes it easier to break the path using standard path manipulation libraries.
What do you think?

nakst: You could ASCII 0x1F Unit Separator. That's unlikely to conflict with any real file paths.

robbert-vdh: It should be easier to parse, which is the motivation behind the suggestion. `:` is a valid character in file paths, but the same notation has been used a lot for referencing specific lines in files. I just checked what grep and co do when file names contain colons:

```
/tmp/grep-party
❯ lt
drwxr-xr-x   - robbert  3 jan 18:37 .
drwxr-xr-x   - robbert  3 jan 18:38 ├── dir-containing-a-:-character
.rw-r--r-- 292 robbert  3 jan 18:38 │  ├── also-containing-a-:-character
.rw-r--r-- 292 robbert  3 jan 18:38 │  ├── some-file
.rw-r--r-- 292 robbert  3 jan 18:38 │  └── trailing-:
.rw-r--r-- 292 robbert  3 jan 18:37 └── file-containing-a-:-character

/tmp/grep-party
❯ grep -R 99
dir-containing-a-:-character/trailing-::99
dir-containing-a-:-character/also-containing-a-:-character:99
dir-containing-a-:-character/some-file:99
file-containing-a-:-character:99

/tmp/grep-party
❯ rg 99
file-containing-a-:-character
99:99

dir-containing-a-:-character/some-file
99:99

dir-containing-a-:-character/trailing-:
99:99

dir-containing-a-:-character/also-containing-a-:-character
99:99

/tmp/grep-party
❯ rg --vimgrep 99
file-containing-a-:-character:99:1:99
dir-containing-a-:-character/some-file:99:1:99
dir-containing-a-:-character/also-containing-a-:-character:99:1:99
dir-containing-a-:-character/trailing-::99:1:99
```

So they don't treat colons differently, and the last colon is the separator. That means that preset IDs cannot contain the separator. So regardless of what separator is used for banks that also needs to be documented. Or if slashes _are_ allowed in preset names, then that should also be documented. So a `bass/gnarly-reese` preset in the `user.dat` bank would be addressed as `file:///home/username/.local/share/awesome-synth/presets/user.dat/bass/gnarly-reese`. In that case the path just always needs to be parsed segment by segment.

I guess I don't have a strong opinion here. Colons might make parsing easier and it at least makes it easier to distinguish bank file paths from preset IDs for humans, but the URI parsing needs to be implemented with care either way.

include/clap/ext/draft/preset-load.h
```diff
@@ -9,19 +9,13 @@ extern "C" {
...
+   // - "plugin://<plugin-id>/<preset-id>"
    //
    // [main-thread]
-   bool(CLAP_ABI *from_container_file)(const clap_plugin_t *plugin,
-                                       const char          *path,
-                                       const char          *preset_id);
+   bool(CLAP_ABI *from_uri)(const clap_plugin_t *plugin, const char *uri);
```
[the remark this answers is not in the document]
robbert-vdh: What if a plugin doesn't have a way to pop up a GUI file dialog while its GUI isn't visible? Or should the plugin always create a standalone window in that case?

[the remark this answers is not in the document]
nakst: Good point. Then I think as you said before, it should always be the host's responsibility. I suggest adding a comment saying something like `// If the preset cannot be loaded for any reason, return false so that the host can process the error as it likes. The plugin should not present error UI/dialogs itself.`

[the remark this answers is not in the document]
robbert-vdh: It might be nice to have some way for the plugin to return a one line error string or something though, since there are a number of potential causes for why the preset could not be loaded.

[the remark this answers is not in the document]
abique: Maybe the plugin can use host_log?
Then the host can simply do:
```C
x = get_last_error_msg_index();
if (!plug->load_from_uri()) {
   y = get_last_error_msg_index();
   // report all the errors from x to y into the error message
}
```

Otherwise the issue is that the host has to pre-allocate the error buffer, which makes the interface a bit more bloated.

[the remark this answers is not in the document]
abique: If the host tries to `use load_from_uri()` and it fails, then it can simply show: "failed to load preset <uri>".
That should be enough for the user, and then he can investigate the engine logs.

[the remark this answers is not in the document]
robbert-vdh: > Otherwise the issue is that the host has to pre-allocate the error buffer, which makes the interface a bit more bloated.

It can be allocated on the stack, so that shouldn't make the API much more cumbersome to use. The error buffer could either have a predefined size, or the function could get a `char*` and a `size_t`/`uint32_t` parameter for the buffer and the buffer's size,and a `size_t*`/`uint32_t*` where the plugin can change the error length to. If that doesn't change from the 0 value the host initialized it to, then there was no error message.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,294 @@
...
+   // binary offset. It's use is entirely up to the plug-in.
+   void(CLAP_ABI *begin_contained_preset)(
+      const struct clap_preset_discovery_metadata_receiver *receiver,
+      const char                                           *path,
+      const char                                           *preset_id);
+
+   // Marks this preset as a bank preset, meaning that it can be assigned to the plug-in as a
+   // preset but will update the banks in the plug-in.
+   void(CLAP_ABI *mark_as_bank_preset)(
+      const struct clap_preset_discovery_metadata_receiver *receiver);
```
Bremmers, line 113: Perhaps this can be replaced by a clap_preset_discovery_flags flag?

abique: Yes it can!

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,293 @@
...
+
+   // Create a preset provider by its id.
+   // The returned pointer must be freed by calling preset_provider->destroy(preset_provider);
+   // The preset provider is not allowed to use the indexer callbacks in the create method.
+   // Returns null in case of error.
+   // [thread-safe]
+   const clap_preset_discovery_provider_t *(CLAP_ABI *create)(
+      const struct clap_preset_discovery_factory *factory,
+      const clap_preset_indexer_t                *indexer,
+      const char                                 *provider_id);
```
robbert-vdh, line 288: Would it make sense to have a global invalidation file? That could be set to, e.g., the plugin's config file or some preset-specific config file. If that changes, then that would cause the host to rescan for added or removed preset providers. Right now there's no way to add or remove a plugin location, and this would enable that.

abique: I think it'd would like for the plugin factory.

robbert-vdh: What do you mean? Or do you mean that we could use the plugin invalidation factory for this in some way?

abique: It is the exact same pattern, so maybe it could be generalized.

include/clap/ext/draft/preset-load.h
abique: One other option is to add a `set_load_uri()` in the metadata receiver, then the plugin is responsible for defining the load uri.

abique: Otherwise yes using `/` forces you to stat the parent until it is a file, but it is OK and only necessary for container presets.

include/clap/ext/draft/preset-load.h
```diff
@@ -2,18 +2,42 @@
...
+                            int32_t              os_error,
+                            const char          *msg);
+
+   // Informs the host that the following preset has been loaded.
+   // This contributes to keep in sync the host preset browser and plugin preset browser.
+   //
+   // [main-thread]
+   void(CLAP_ABI *loaded)(const clap_plugin_t *plugin, const char *uri);
+} clap_host_preset_load_t;
+
```
Bremmers, line 40: Both functions should have "const clap_host_t *host" instead of "const clap_plugin_t *plugin"

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,310 @@
...
+   void(CLAP_ABI *set_invalidation_watch_file)(const struct clap_preset_discovery_indexer *indexer,
+                                               const char                                 *path);
+
+   // Query an extension.
+   // The returned pointer is owned by the indexer.
+   // It is forbidden to call it before provider->init().
+   // You can call it within provider->init() call, and after.
+   const void *(CLAP_ABI *get_extension)(const struct clap_preset_discovery_provider *provider,
+                                         const char                                  *extension_id);
+} clap_preset_indexer_t;
```
Bremmers, line 277: Should be clap_preset_discovery_indexer_t I think?
Also appears in clap_preset_discovery_factory.create()

include/clap/ext/draft/preset-load.h
abique: Good catch! :-)

include/clap/factory/draft/preset-discovery.h
abique: Good catch again! :-)

Bremmers: Haha, you've found a bug yourself, and fixed it :-)

I meant this:

line 277 should be:
} clap_preset_discovery_indexer_t;

line 304 should be:
const clap_preset_discovery_indexer_t                *indexer,

abique: Thank you again!

Bremmers (approved):

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,310 @@
...
+   // Examples:
+   // kick, drum, tom, snare, clap, cymbal, bass, lead, metalic, hardsync, crossmod, acid,
+   // distorted, drone, pad, dirty, etc...
+   void(CLAP_ABI *add_feature)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                               const char                                           *feature);
+
+   // Adds extra information to the metadata.
+   void(CLAP_ABI *add_extra_info)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                  const char                                           *key,
+                                  const char                                           *value);
```
Trinitou, line 165: Could this be used e.g. to add information about a preview sample that could be played back when selecting a preset in the host plugin browser?

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,310 @@
...
+   // The timestamps are in seconds since EPOCH.
+   // If one of the time isn't known, then set it to 0.
+   // If this function is not called, then the indexer may look at the file's creation and
+   // modification time.
+   void(CLAP_ABI *set_timestamps)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                  uint64_t creation_time,
+                                  uint64_t modification_time);
+
+   // Adds a feature to the preset.
+   // See plugin-features.h and preset-features.h
```
Trinitou, line 151: Where do I find preset-features.h? Or is it yet to come?

include/clap/factory/draft/preset-discovery.h
abique: Yes it could, for now there isn't such thing.
I believe that having a preview should be as simple as loading the preset and sending a note or some audio for effects, this would save disk space.

include/clap/factory/draft/preset-discovery.h
abique: preset feature was removed, because it wasn't complete enough
I will update the header, good catch!

ChangeLog.md
```diff
@@ -1,3 +1,19 @@
+# Changes in 1.1.7
+
+* Add a [factory](incl/../include/clap/factory) folder for better organization and move our factories there
```
robbert-vdh, line 3: The path to the directory is incorrect.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,315 @@
...
+   1. clap_plugin_entry.get_factory(CLAP_PRESET_DISCOVERY_FACTORY_ID)
+   2. clap_preset_discovery_factory_t.create(...)
+   3. clap_preset_discovery_provider.init() (only necessary the first time, declarations
+   can be cached)
+        `-> clap_preset_discovery_indexer.declare_filetype()
+        `-> clap_preset_discovery_indexer.declare_location()
+        `-> clap_preset_discovery_indexer.declare_collection() (optional)
+        `-> clap_preset_discorery_indexer.set_invalidation_watch_file() (optional)
+   4. crawl the given locations and monitor file system changes
+        `-> clap_preset_discovery_indexer.get_metadata() for each presets files
```
robbert-vdh, line 22: Bit of a radical suggestion, but would adding a `clap_plugin_` and `clap_host_` prefix to `clap_preset_discovery_provider`, and `clap_preset_discovery_indexer`, and `clap_preset_discovery_metadata_receiver` introduce too much clutter? I always forget which side uses which, and I'm sure anyone else first looking at this extension will run into the same issue. Consistent naming conventions might make this a bit simpler to grasp.

Also, this reads clap\_preset\_**discorery**\_indexer right now. :wink:

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,315 @@
...
+   // Adds a creator name for the preset.
+   void(CLAP_ABI *add_creator)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                               const char                                           *creator);
+
+   // Sets a description of the preset.
+   void(CLAP_ABI *set_description)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                   const char *description);
+
+   // Sets the creation time and last modification time of the preset.
+   // The timestamps are in seconds since EPOCH.
```
robbert-vdh, line 144: Maybe just explicitly say 'since the Unix epoch', since there technically are multiple reference epochs used in computing. Also, even though this is unlikely to result in issues in practice, it might be a good idea to specify whether this should be in UTC or in the user's locale timezone. Otherwise plugins/hosts will mix this up, and different platforms have different conventions.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,315 @@
...
+   void(CLAP_ABI *add_extra_info)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                  const char                                           *key,
+                                  const char                                           *value);
+} clap_preset_discovery_metadata_receiver_t;
+
+typedef struct clap_preset_discovery_filetype {
+   const char *name;
+   const char *description;
+
+   // `.' isn't included in the string.
```
robbert-vdh, line 173: But is it still allowed? Say my preset files are in the format `preset-name.json.zstd`, can I still use `json.zstd` here?

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,315 @@
...
+} clap_preset_discovery_filetype_t;
+
+// Defines a place in which to search for presets
+typedef struct clap_preset_discovery_location {
+   uint32_t    flags; // see enum clap_preset_discovery_flags
+   const char *name;  // name of this location
+
+   // URI
+   // - file:/// for pointing to a file or directory; directories are scanned recursively
+   //   eg: file:///home/abique/.u-he/Diva/presets/
```
robbert-vdh, line 185: It may also be a good idea to list an example of a path on Windows here since things work a little differently there, and different hosts/plugins may end up doing different things otherwise. So this would become `file://C:/Users/abique/<whatever the equivalent path is on Windows>`.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,315 @@
...
+   // preset using the OS features (file system monitoring).
+   //
+   // The invalidation file is useful if:
+   // - the set of filetypes changes
+   // - the set of locations changes
+   // - the set of collections changes
+   // - the metadata extraction code changes is located outside of the DSO containing this
+   //   preset provider
+   void(CLAP_ABI *set_invalidation_watch_file)(const struct clap_preset_discovery_indexer *indexer,
+                                               const char                                 *path);
```
robbert-vdh, line 274: Can these four functions be called before `clap_preset_discovery_provider::init()` has been called? I'd assume not but it never mentions this anywhere.

robbert-vdh (changes requested):

include/clap/factory/draft/preset-discovery.h
Bremmers: https://en.wikipedia.org/wiki/File_URI_scheme says the Shell API works like this (file:/// rather than file://):

file:///c:/Users/abique/.u-he/Diva/presets/

Trinitou:
Maybe the factories concept could be documented in the readme as well. Following aspects might be interesting for developpers looking into CLAP for the first time:

- list of fundamental factories / extra / third party factories (analog to the extension lists)
  - plugin_factory is the one necessary for creating plugin instances
    - (one .clap file can "contain" multiple plugins)
  - ARA: third party extension/factory -> [ARA API CLAP factory & extension](https://github.com/Celemony/ARA_API/blob/7ba766334921477e14984c696f5217ecb757d511/ARACLAP.h) (under third-party factories/extensions)
- extensions apply per plugin instance whereas factories are "global"/singleton/per .clap file

I'm not 100% sure about the exact formulations.

Also it could be good to not make it a new "Factories" section in the Readme.md but integrate the points about factories into the "Extensions" section
- maybe then rename the "Extensions" section to "Factories & Extensions"
- As there aren't so many factories, maybe it would be OK to list them together with extensions but tag them as `(Factory)`
- e.g. under "Fundamental extensions" add a bullet point `- Plugin factory (Factory): lets the host create plugin instances`

include/clap/factory/draft/preset-discovery.h
abique: I've added a reference to C's `time()` function, so it is the number of seconds since EPOCH which is always UTC right?

include/clap/factory/draft/preset-discovery.h
abique: Yes, it is similar to how works `/etc/mime.types`.

include/clap/factory/draft/preset-discovery.h
abique: No they can't, I've added a comment in `create()`.

include/clap/factory/draft/preset-discovery.h
abique: Fixed.

include/clap/factory/draft/preset-discovery.h
robbert-vdh: It is yeah. Probably doesn't hurt to specifically mention it, but I think most APIs for getting file modification times are relative to UTC. I checked Windows' filetime API and that also uses UTC, albeit with a different epoch compared to everything else.

include/clap/factory/draft/preset-discovery.h
Trinitou: > Bit of a radical suggestion, but would adding a `clap_plugin_` and `clap_host_` prefix to `clap_preset_discovery_provider`, and `clap_preset_discovery_indexer`, and `clap_preset_discovery_metadata_receiver` introduce too much clutter?

I also would vote against host/plugin prefixes for factory methods. As extensions are per plugin (instance), the prefixes make sense for extension methods. Factories are per .clap file and don't necessarily apply to a single plugin. (Also it isn't given that there already is a clap host when calling a factory method. Only when the plugin instances are actually created, the host is needed.)

Trinitou: @robbert-vdh Yeah, I think you are right (as well?). When talking about host/plugin I meant the `clap_host_t`/`clap_plugin_t` structs. I just think that it's easy to confuse both host/plugin levels ATM. And those prefixes would confuse it even more.

include/clap/ext/draft/preset-load.h
```diff
@@ -32,9 +32,11 @@ typedef struct clap_host_preset_load {
 
    // Informs the host that the following preset has been loaded.
    // This contributes to keep in sync the host preset browser and plugin preset browser.
+   // If the preset was loaded from a container file, then the subpath must be set, otherwise it
+   // must be null.
    //
    // [main-thread]
-   void(CLAP_ABI *loaded)(const clap_host_t *host, const char *uri);
+   void(CLAP_ABI *loaded)(const clap_host_t *host, const char *uri, const char *subpath);
```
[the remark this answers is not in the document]
abique: Both options were possible, though subpath is human friendly, so maybe better for debugging, while load_key is better for loading the preset as it isn't meant for human display.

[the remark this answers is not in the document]
Bremmers: But the function is intended to let the host keep track of which preset is loaded. That's a code thing, not a human readable thing.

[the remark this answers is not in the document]
Trinitou: Aren't we all a code thing?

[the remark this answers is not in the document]
abique: OK

include/clap/ext/draft/preset-load.h
```diff
@@ -2,18 +2,43 @@
...
-   // Loads a preset in the plugin native preset file format from a path.
+   // Loads a preset in the plugin native preset file format from a URI. eg:
+   // - "file:///home/abique/.u-he/Diva/Presets/Diva/HS Bass Nine.h2p", load_key: null
+   // - "plugin://<plugin-id>", load_key: <XXX>
+   //
+   // The preset discovery provider defines the uri and load_key to be passed to this function.
+   //
    // [main-thread]
-   bool(CLAP_ABI *from_file)(const clap_plugin_t *plugin, const char *path);
+   bool(CLAP_ABI *from_uri)(const clap_plugin_t *plugin, const char *uri, const char *load_key);
```
nakst, line 19: Is this call synchronous? e.g. if the plugin needs to load large sample files from the user's drive for the preset, do these loads need to finish before the `from_uri` call returns?

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,321 @@
...
+   uint64_t    flags;        // see enum clap_preset_discovery_flags
+   const char *id;           // sound pack identifier
+   const char *name;         // name of this sound pack
+   const char *description;  // reasonably short description of the sound pack
+   const char *homepage_url; // url to the pack's homepage
+   const char *vendor;       // sound pack's vendor
+   const char *image_uri;    // may be an image on disk or from an http server
+
+   // release date, in numbe of seconds since UNIX EPOCH, 0 if unavailable
+   uint64_t release_timestamp;
```
Trinitou, line 197: This made me think whether it might be helpful to have a version number for soundpacks (?) Probably useful for soundpack updates.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,321 @@
...
+                                   const char *description);
+
+   // Sets the creation time and last modification time of the preset.
+   // The timestamps are in seconds since UNIX EPOCH, see C's time_t time(time_t *).
+   // If one of the time isn't known, then set it to 0.
+   // If this function is not called, then the indexer may look at the file's creation and
+   // modification time.
+   void(CLAP_ABI *set_timestamps)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                  uint64_t creation_time,
+                                  uint64_t modification_time);
```
Trinitou, line 145: Maybe we could introduce a new 'clap_timestamp' type for these UNIX EPOCH timestamps? Then the UNIX EPOCH explanation could be put next to this type and wouldn't have to be mentioned everywhere.
Or maybe call the type 'clap_unix_epoch_timestamp' so that the correct usage is clarified immediately.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,321 @@
...
+   // Sets a description of the preset.
+   void(CLAP_ABI *set_description)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                   const char *description);
+
+   // Sets the creation time and last modification time of the preset.
+   // The timestamps are in seconds since UNIX EPOCH, see C's time_t time(time_t *).
+   // If one of the time isn't known, then set it to 0.
+   // If this function is not called, then the indexer may look at the file's creation and
+   // modification time.
+   void(CLAP_ABI *set_timestamps)(const struct clap_preset_discovery_metadata_receiver *receiver,
```
Trinitou, line 143: Thinking about preset versioning/variations (see also my loose thoughts/ideas #284 ): Could we allow multiple presets with same subpath and creation timestamps but different modification timestamps? This might be helpful for managing different versions/variations of the same preset.

include/clap/factory/draft/preset-discovery.h
defiantnerd: I‘d like that, maybe with helper functions

include/clap/factory/draft/preset-discovery.h
Trinitou: It could then be up to the host to decide whether to show only the newest version of a preset, list the preset variations next to each other or to offer a separate selector for the preset versions (like the selector when someone edited a github comment ;) ).

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,321 @@
...
+typedef struct clap_preset_discovery_soundpack {
+   uint64_t    flags;        // see enum clap_preset_discovery_flags
+   const char *id;           // sound pack identifier
+   const char *name;         // name of this sound pack
+   const char *description;  // reasonably short description of the sound pack
+   const char *homepage_url; // url to the pack's homepage
+   const char *vendor;       // sound pack's vendor
+   const char *image_uri;    // may be an image on disk or from an http server
+
+   // release date, in numbe of seconds since UNIX EPOCH, 0 if unavailable
```
Trinitou, line 196:
```suggestion
   // release date, in number of seconds since UNIX EPOCH, 0 if unavailable
```

include/clap/ext/draft/preset-load.h
robbert-vdh: Presumably the semantics are exactly the same as `clap_plugin_state::load()`.

nakst: The state extension doesn't document it either way. For now, I've made a mini-extension just for my internal tools (I needed it for unit testing)
```
struct clap_plugin_nakst_ensure_samples_loaded { void (*ensure_samples_loaded)(const clap_plugin_t *plugin); };
#define CLAP_EXT_NAKST_ENSURE_SAMPLES_LOADED "nakst.ensure_samples_loaded"
```

abique: Yes this call is supposed to be synchronous, you can still decide to return quickly and finish loading in background if your code supports it.
I think that in  your situation: "loading samples", you can block in `activate()` if the background sample loading isn't finished.

abique: If the plugin is already activated, then it is a best effort loading, as if you click on the preset from your GUI.

include/clap/factory/draft/preset-discovery.h
abique: What would the host do with that?
Release timestamp could be used in a search query like "All soundpacks released in 2003".
But would the user search for "all soundpacks at fourth update level"? I doubt it.

include/clap/factory/draft/preset-discovery.h
abique: Yes good idea!

include/clap/factory/draft/preset-discovery.h
abique: It'd be too complicated.
That would make the API more complicated for a feature very few will use.
If you make multiple variations, then I'd suggest to use similar preset name and put the variation at the end.
Though I suggest that when making a sound pack, to just keep the best variation so it is then less content to browse for the user. The user's time is precious and it is better to not make him try 5 times the same preset, also to not unnecessarily clutter the preset browser.

Trinitou (commented):
We could bring back some info about unknown/unavailable timestamps using the new CLAP_TIMESTAMP_UNKNOWN constant

abique:
> We could bring back some info about unknown/unavailable timestamps using the new CLAP_TIMESTAMP_UNKNOWN constant

Thanks! :)

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,332 @@
...
+   // the plugin wants but it could also be some other unique id like a database primary key or a
+   // binary offset. It's use is entirely up to the plug-in.
+   //
+   // If the function returns false, the the provider must stop calling back into the receiver.
+   bool(CLAP_ABI *begin_preset)(const struct clap_preset_discovery_metadata_receiver *receiver,
+                                const char                                           *name,
+                                const char                                           *load_key);
+
+   // Adds a plug-in id that this preset can be used with.
+   // plugin_abi: 0 for CLAP
```
Bremmers, line 127: plugin_id.abi: 'clap' for CLAP

abique: Good catch!
Fixed.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -0,0 +1,318 @@
...
+   uint32_t    flags; // see enum clap_preset_discovery_flags
+   const char *name;  // name of this location
+
+   // URI:
+   // - file:/// for pointing to a file or directory; directories are scanned recursively
+   //   eg: file:///home/abique/.u-he/Diva/Presets/Diva (on Linux)
+   //   eg: file:///C:/Users/abique/Documents/u-he/Diva.data/presets/ (on Windows)
+   //
+   // - plugin://<resource-id> for presets which are bundled inside the plugin itself.
+   //   The resource-id string is defined by the plugin and should unique to avoid clashing
```
Trinitou, line 197:
```suggestion
   //   The resource-id string is defined by the plugin and should be unique to avoid clashing
```

defiantnerd (approved):
ok for me.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -193,10 +193,8 @@ typedef struct clap_preset_discovery_location {
    //   eg: file:///home/abique/.u-he/Diva/Presets/Diva (on Linux)
    //   eg: file:///C:/Users/abique/Documents/u-he/Diva.data/presets/ (on Windows)
    //
-   // - plugin://<resource-id> for presets which are bundled inside the plugin itself.
-   //   The resource-id string is defined by the plugin and should be unique to avoid clashing
-   //   with other providers, because when preset_load.from_uri() will be called, the
-   //   provider's origin won't be known.
+   // - plugin:// for presets which are bundled within the plugin DSO.
+   //   The path component may be empty.
```
robbert-vdh, line 197: It may seem obvious but just to prevent future hosts from implementing it incorrectly it wrong it might be worth explicitly mentioning that the `plugin://[path]` URIs are local to this specific .clap plugin, and that multiple plugins may use the same `plugin://` URI.

abique: Do you have a sentence to suggest?
I could also say that the path component must be empty.

abique: I'll force it to be exactly `plugin://` that simplifies everything.

robbert-vdh: One empty path per plugin should also work. The only use cases I can think of where exposing multiple internal preset locations would be useful are for grouping built in presets into multiple sound packs, and for providing internal preset data for plugins using other plugin APIs (but that seems like both a fringe use case, and something that wouldn't consistently work in the first place since you may lack up to date information on that plugin's internal presets).

abique: Maybe we just go with the simpler approach and if we need more, then we improve it.

PR #283, opened 2023-01-11
added  CLAP_EXT_AUDIO_PORTS_CONFIG_INFO extension

defiantnerd:
added possibility to inquiry the exact bus layout for an configuration.

### DIFF
--- a/include/clap/ext/audio-ports-config.h
+++ b/include/clap/ext/audio-ports-config.h
@@ -2,6 +2,7 @@
 
 #include "../string-sizes.h"
 #include "../plugin.h"
+#include "audio-ports.h"
 
 /// @page Audio Ports Config
 ///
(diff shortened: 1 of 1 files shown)


abique:
You need to update the surround and ambisonic extensions too.

abique:
Please could you also add the changelog entry?

include/clap/ext/audio-ports-config.h
```diff
@@ -72,6 +72,13 @@ typedef struct clap_plugin_audio_ports_config {
 
 typedef struct clap_plugin_audio_ports_config_info {
   // get info about about an audio port.
+
+  // gets the clap_id of the currently selected config
+  // clap_id might now match the index, can be used to call
+  // clap_plugin_audio_ports_config::select()
+  // [main-thread]
+  clap_id(CLAP_ABI* current_config)(const clap_plugin_t* plugin);
```
robbert-vdh, line 80: Is this needed? The host is the one selecting a config. So it knows the current config.

defiantnerd: not necessarily - the plugin itself could also switch configurations (like in a user selection, as the commentary states) and the host would not know which one is selected now.

PR #291, opened 2023-01-29
1.1.8

abique:


include/clap/ext/draft/preset-load.h
```diff
@@ -16,7 +16,10 @@ typedef struct clap_plugin_preset_load {
    // The preset discovery provider defines the uri and load_key to be passed to this function.
```
Bremmers, line 16: "uri" should now be "location"

abique: Good catch!
I also fixed a few other place where we used URI instead

include/clap/ext/draft/preset-load.h
```diff
@@ -26,7 +26,8 @@ typedef struct clap_host_preset_load {
    //
    // [main-thread]
    void(CLAP_ABI *on_error)(const clap_host_t *host,
-                            const char        *uri,
+                            uint32_t           location_kind,
+                            const char        *location,
                             int32_t            os_error,
                             const char        *msg);
```
Bremmers, line 33: There should be a load_key parameter as well I think?

abique: Yes.

include/clap/ext/draft/configurable-audio-ports.h
```diff
@@ -0,0 +1,52 @@
...
+   // [main-thread && !active]
+   bool(CLAP_ABI *request_configuration)(
+      const clap_plugin_t                                *plugin,
+      const struct clap_audio_port_configuration_request *requests,
+      uint32_t                                            request_count);
+} clap_plugin_configurable_audio_ports_t;
+
+#ifdef __cplusplus
+}
+#endif
```
robbert-vdh, line 52: Missing linefeed at the end of the file.

include/clap/ext/draft/extensible-audio-ports.h
```diff
@@ -0,0 +1,32 @@
...
+
+   // Asks the plugin to remove a port.
+   // Returns true on success.
+   // [main-thread && !is_active]
+   bool(CLAP_ABI *remove_port)(const clap_plugin_t *plugin, bool is_input, uint32_t index);
+} clap_plugin_extensible_audio_ports_t;
+
+#ifdef __cplusplus
+}
+#endif
```
robbert-vdh, line 32: Also missing a linefeed here.

include/clap/ext/draft/preset-load.h
```diff
@@ -9,14 +9,14 @@ extern "C" {
 #endif
```
robbert-vdh, line 9: The draft extension version needs to be incremented.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -185,28 +193,21 @@ typedef struct clap_preset_discovery_filetype {
...
-   //   eg: file:///home/abique/.u-he/Diva/Presets/Diva (on Linux)
-   //   eg: file:///C:/Users/abique/Documents/u-he/Diva.data/Presets/ (on Windows)
-   //
-   // - plugin:// for presets which are bundled within the plugin DSO.
-   //   In that case, the uri must be exactly `plugin://` and nothing more.
-   const char *uri;
+   uint32_t    flags;    // see enum clap_preset_discovery_flags
+   const char *name;     // name of this location
+   uint32_t    kind;     // See clap_preset_discovery_location_kind
+   const char *location; // Actual location in which to crawl presets
```
robbert-vdh, line 199: So, any mandates on the use of slashes on Windows? Both regular slashes and backslashes should be handled by the host? Probably a good idea to document that regardless, even thought it's unlikely to cause problems unless people implement their own file APIs.

robbert-vdh (changes requested):

include/clap/factory/draft/preset-discovery.h
abique: I think if the OS functions allows '/' which I think they do, then '/' are allowed.

abique: I will elaborate on the location attr

include/clap/factory/draft/preset-discovery.h
```diff
@@ -196,7 +199,11 @@ typedef struct clap_preset_discovery_location {
    uint32_t    flags;    // see enum clap_preset_discovery_flags
    const char *name;     // name of this location
    uint32_t    kind;     // See clap_preset_discovery_location_kind
-   const char *location; // Actual location in which to crawl presets
+
+   // Actual location in which to crawl presets.
+   // For FILE kind, the location can be either a path to a directory or a file.
+   // For PLUGIN kind, the location is arbitrary.
```
robbert-vdh, line 205: Does arbitrary mean that the parameter should be ignored, or that the host should repeat the value from the `location` field verbatim?

abique: The current idea is that the provider can use whatever it wants for the location, and then the indexer/host just give that string back.

I think it can lead to duplicates, if the plugin is installed twice with and two providers uses different location for the same plugin target.

While if the plugin is the preset container, then the location should be irrelevant. I'm very tempted to force it to be null.

abique: Let's force it to null for now, and if anyone needs it then we re-visit it.

robbert-vdh: > The current idea is that the provider can use whatever it wants for the location, and then the indexer/host just give that string back.

Yeah that makes sense to me, probably worth explicitly mentioning that the arbitrary string should still be used by the host and not treated like arbitrary data that can be ignored. I don't think allowing locations can lead to more duplicates than not allowing locations though. The location is still relative to the plugin's shared library.

abique: Yeah but if the benefit is 100% hypothetical, then let's get rid of it.

abique: I force the null location.

include/clap/factory/draft/preset-discovery.h
```diff
@@ -48,12 +48,24 @@
...
+   // The preset are located in a file on the OS filesystem.
+   // The location is then a path which works with the OS file system functions (open, stat, ...)
+   // So both '/' and '\' shall work on Windows as a separator.
+   CLAP_PRESET_DISCOVERY_LOCATION_FILE = 0,
+
+   // The preset is bundled within the plugin DSO itself.
+   // The location must then be null, as the preset are within the plugin itsel and then the plugin
+   // will act as a preset container.
+   CLAP_PRESET_DISCOVERY_LOCATION_PLUGIN = 1,
+};
```
Trinitou, line 67: When reading the documentation, I think that it might be better to have two location struct types:

- `clap_preset_discovery_file_location_t`
- `clap_preset_discovery_plugin_location_t`

The `kind` member wouldn't be needed anymore.
The `clap_preset_discovery_plugin_location_t` would be without the `location` member.
For declaration there could be two separate methods:

- `declare_file_location`
- `declare_plugin_location`

I think it would prevent some of the clutter of the current documentation.

Trinitou (changes requested):
The second question would be concerning the `clap_preset_discovery_metadata_receiver::add_plugin_id` method:

Should it be possible to add plugin id's that aren't provided by the .clap file itself? I mean, in theory now one could write a clap that only contains the preset discovery factory and provides presets for other clap plugins. Or should that be forbidden?

include/clap/factory/draft/preset-discovery.h
abique: Yes you can write a .clap file which only contains a preset provider, and targets other plugins.

abique: We can have additional location kind in the future maybe.
This needs to be flexible.
We could still go for a struct approach, but it'd be something like the events where you need a cast.

Trinitou: > Yes you can write a .clap file which only contains a preset provider, and targets other plugins.

Ok, understand.

> We can have additional location kind in the future maybe.

Yes, maybe online preset resources.

> This needs to be flexible. We could still go for a struct approach, but it'd be something like the events where you need a cast.

Understand as well. Conversation resolved, I'd say ;)

Bremmers (approved):

ChangeLog.md
```diff
@@ -1,3 +1,23 @@
...
+* [configurable-audio-ports.h](include/clap/ext/draft/configurable-audio-ports.h): new exetension which lets the host **push** an audio ports configuration request, this leads to a simpler workflow for surround host and plugins
+* [surround.h](include/clap/ext/draft/surround.h):
+  * remove `get_preferred_channel_map()` in favor of the push approach via [configurable-audio-ports.h](include/clap/ext/draft/configurable-audio-ports.h)
+  * remove `config_id` argument from `get_info()`
+* [ambisonic.h](include/clap/ext/draft/ambisonic.h): remove `config_id` argument from `get_info()`
+* [preset-load.h](include/clap/ext/draft/preset-load.h): use a location_kind + location approach instead of URI
+
+## Draft factories
+
+* [preset-discovery.h](include/clap/factory/draft/preset-discovery.h): use a location_kind + location approach instead of URI
```
robbert-vdh, line 19: Some other changes are missing here:

```suggestion
* [preset-discovery.h](include/clap/factory/draft/preset-discovery.h): 
  * use a location_kind + location approach instead of URI
  * document which descriptor fields are optional
  * allow optional preset names in the metadata for non-container presets
```

include/clap/ext/draft/configurable-audio-ports.h
```diff
@@ -0,0 +1,58 @@
...
+#include "../audio-ports.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// This extension lets the host configure the plugin's input and output audio ports.
+// This is a "push" approach to audio ports configuration.
+static CLAP_CONSTEXPR const char CLAP_EXT_CONFIGURABLE_AUDIO_PORTS[] =
+   "clap.configurable-audio-ports.draft0";
```
robbert-vdh, line 12: Is `clap.audio-ports-config-info/draft-0` still needed with this extension existing? They technically cover different use cases, but `audio-ports-config-info` was added to better accommodate the VST3 bus arrangement setting protocol, and this new extension directly implements that protocol. If there's no demand for it anymore than it may be worth getting rid of. The changes to the ambisonics and surround extensions that accompanied this extension have also been removed.

robbert-vdh (changes requested):
Went through it one more time, spotted a couple more things. Otherwise LGTM!

include/clap/ext/draft/configurable-audio-ports.h
abique: I also think we can remove it. @defiantnerd Do you agree?

ChangeLog.md
```diff
@@ -1,3 +1,26 @@
+# Changes in 1.1.8
+
+* [params.h](include/clap/ext/params.h): document how persisting parameter values between sessions should be implemented
+* [state.h](include/clap/ext/state.h): add basic documentation regarding what state should be saved and how plugins should interact with buffers
+* various documentation fixes (essentially typos)
+
+## Draft extensions
+
+* [extensible-audio-ports.h](include/clap/ext/draft/extensible-audio-ports.h): new extension which lets the host add ports to a plugin
+* [configurable-audio-ports.h](include/clap/ext/draft/configurable-audio-ports.h): new exetension which lets the host **push** an audio ports configuration request, this leads to a simpler workflow for surround host and plugins
```
Trinitou, line 10:
```suggestion
* [configurable-audio-ports.h](include/clap/ext/draft/configurable-audio-ports.h): new extension which lets the host **push** an audio ports configuration request, this leads to a simpler workflow for surround host and plugins
```

robbert-vdh: If you're going to correct the typos, then change it to 

```suggestion
* [configurable-audio-ports.h](include/clap/ext/draft/configurable-audio-ports.h): new extension allowing the host to **push** an audio ports configuration request, resulting in a simpler workflow for surround host and plugins
```

:wink: 

Or 'that lets'. 'Which' is not correct here, and the 'this leads' part either needs to be its own sentence or be rewritten without 'this'.

include/clap/ext/state.h
```diff
@@ -3,6 +3,24 @@
...
+/// ## Notes on using streams
+///
+/// When working with `clap_istream` and `clap_ostream` objects to load and save
+/// state, it is important to keep in mind that the host may limit the number of
+/// bytes that can be read or written at a time. The return values for the
+/// stream read and write functions indicate how many bytes were actually read
+/// or written. You need to use a loop to ensure that you read or write the
+/// entirety of your state. Don't forget to also consider the negative return
+/// values for the end of file and IO error codes.
+
```
Trinitou, line 23: see #312

include/clap/ext/draft/configurable-audio-ports.h
abique: Let's continue this one in 1.1.9 so we allow some time.

robbert-vdh (approved):

defiantnerd (approved):
regarding the audio busses, we can continue on 1.1.9, for 1.1.8 approved.

baconpaul (approved):
Happy to approve 

some comments for 1.9

optional const char* should be set to nullptr or a zero length string? I assume nullptr but let’s say this. 

We are inconsistent in draft names. Foo.draft0 .draft-0 .draft/0. We should pick one

but these are both 1.9 things good to merge

PR #294, opened 2023-02-01
Document the relationship between the params and state extensions

robbert-vdh:
As discussed. Making this behavior explicit is necessary because some plugins are already relying on host-specific implementation details.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,8 @@
+# Changes in 1.1.8
+
+* [params.h](include/clap/ext/params.h): document how persisting parameter values between sessions should be implemented
+* [state.h](include/clap/ext/state.h): add basic documentation regarding what state should be saved and how plugins should interact with buffers
+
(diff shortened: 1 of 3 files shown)


baconpaul:
I agree with these changes.

robbert-vdh:
The motivation behind forbidding manually persisting parameter values (or at least kindly asking hosts not to) is that unless it is mandatory, plugins can't rely on this behavior and need to implement state anyways (and they'd lose their parameter values if they need to implement state later, but that's beside the point). And if every plugin with parameters thus needs to implement the state extension, then there's no reason for hosts to have this manual persisting behavior. So in the end it's either not necessary or a source of potential confusion when plugins do start relying on one host's behavior only to the realize that their plugin doesn't work as expected with most other hosts.

Bremmers:
I agree with this too!

abique:
@robbert-vdh excellent work!

Issue #313, opened 2023-03-14
GUI extension - some methods are missing the floating/!floating tag


Trinitou:
in `gui.h`:
Some methods don't have the floating / !floating tag although they are only for one of the modes, e.g
- can_resize
- adjust_size
- set_size
- request_resize?

Instead it is mentioned in the comment that the method is only for embedded windows. I think only the tags should be used instead of the comments.

abique:
Maybe you want to make a PR for that?

Trinitou:
> Maybe you want to make a PR for that?

If i find the time, I'll do. If somebody else is faster, no problem 😉

Trinitou:
> Maybe you want to make a PR for that?

-> #315

PR #315, opened 2023-03-15
GUI extension: !floating tag

Trinitou:
PR for issue #313 

Concerning `get_resize_hints` and `resize_hints_changed`:

- Probably for another issue/PR: I just wondered that those methods are not mentioned in the rest of the documentation. When just reading the comments next to the methods themselves, I don't know when and why they get called.
- But I hope I got it right to assign the !floating tag to them as well ❓


abique:
Thank you :+1:

PR #352, opened 2023-10-07
midi controls extension

Bremmers:
midi-info extension


include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,84 @@
...
+
+typedef struct clap_plugin_midiinfo {
+   // Returns the number of clap_midiinfo structs for a channel.
+   // channel is 0..15
+   // [main-thread]
+   uint32_t(CLAP_ABI *count)(const clap_plugin_t *plugin,
+                             uint32_t             channel);
+
+   // Fills midiinfo. Returns true on success.
+   // Important: the most important controls (from a performing musician's point of view) should be listed first,
```
Trinitou, line 63: This is not a requirement but a recommendation. I think if it is important, it should somehow be s requirement otherwise this sentence could be left out. To make it a requirement there should be a technical way to test whether it was met.

Bremmers: clap_midiinfo could have a 'priority' integer member.

The MIDI-CI Property Exchange "ChCtrlList" resource has an (optional) Priority property that can be 1..5. On top of that the controls should be ordered from important to unimportant. We don't have to copy this of course. We could have a priority from 0 to 1000 and ditch the ordering requirement. The latter is probably cumbersome for plugins anyway.

Tagging the important controls in some way is a key requirement: the important ones will be automatically mapped to hardware knobs. So if a plugin gets this wrong the user ends up with useless knobs.

include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,84 @@
...
+
+   //CLAP note expressions.
+   //A host may map MIDI messages to note expressions, so it needs to know which note expressions are supported.
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_VOLUME = 0x100,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_PAN = 0x101,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_TUNING = 0x102,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_VIBRATO = 0x103,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_EXPRESSION = 0x104,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_BRIGHTNESS = 0x105,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_PRESSURE = 0x106
```
baconpaul, line 38: wouldn't this expression of which NE you support be better in a different extension? We've replicated the same thing into the VST wrapper so seeing it here makes me think a 'note expression capabilities' separate extension may be more applicable?

include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include "../../plugin.h"
+
+// This extension lets a plugin expose it's MIDI implementation.
+// The primary goal is to allow a host to create a MIDI-CI Property Exchange "ChCtrlList" resource for the plugin,
+// so keyboard controllers can assign their knobs to suitable controllers.
+
+//TODO : clap_midiinfo could contain more (optional) info (min/max, stepcount, paramPath etc.)
```
baconpaul, line 9: This information is all in the param; and this extension won't work if you don't have param. So can you just glean it from there?

include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include "../../plugin.h"
+
+// This extension lets a plugin expose it's MIDI implementation.
+// The primary goal is to allow a host to create a MIDI-CI Property Exchange "ChCtrlList" resource for the plugin,
+// so keyboard controllers can assign their knobs to suitable controllers.
+
+//TODO : clap_midiinfo could contain more (optional) info (min/max, stepcount, paramPath etc.)
+//TODO : for completeness this should work per-input port. Is that worth it?
```
baconpaul, line 10: yes. it is easier to add port now than have to retrofit it later, and all our other APIs have port.

include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,84 @@
...
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_VIBRATO = 0x103,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_EXPRESSION = 0x104,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_BRIGHTNESS = 0x105,
+   CLAP_MIDIINFO_STATUS_NOTE_EXPRESSION_PRESSURE = 0x106
+};
+
+/* This describes a MIDI control/message */
+typedef struct clap_midiinfo {
+   //CLAP_MIDIINFO_STATUS, describes what kind of control this struct is about.
+   status: uint32_t;
```
baconpaul, line 44: This syntax is backwards from the syntax we use everywhere else, namely `uint32_t status`. I didn't even know the postfix type worked with C! but I think we should be consistent.

include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,84 @@
...
+typedef struct clap_midiinfo {
+   //CLAP_MIDIINFO_STATUS, describes what kind of control this struct is about.
+   status: uint32_t;        
+
+   //controller number, nrpn number etc.
+   //zero if not applicable the status value
+   number: uint32_t;
+
+   //Display name
+   //If name is empty the host can assume the standard MIDI name
```
baconpaul, line 51: What is the 'standard MIDI name'? Is it determined by controller number?

include/clap/ext/draft/midi-info.h
baconpaul: Tagging @defiantnerd on this comment too.

include/clap/ext/draft/midi-info.h
Bremmers: Oops, that's Delphi code :-)
Yet github seems to be able to compile it. Odd.

baconpaul: well this .h file isn't included anywhere is it? so I'm not sure the compiler even ran it.

include/clap/ext/draft/midi-info.h
Bremmers: This extension isn't related to params at all I think. Why do you think so?

include/clap/ext/draft/midi-info.h
baconpaul: Yeah unless you are included <https://github.com/free-audio/clap/blob/41964fa3f44fdbf6feb20df528a9cdf2cc3b05ba/include/clap/clap.h#L64> here you won't get compiled by CI.

include/clap/ext/draft/midi-info.h
Bremmers: Would be fine with me.

include/clap/ext/draft/midi-info.h
baconpaul: Specifically what's the relationship between this extension and

https://github.com/free-audio/clap/blob/main/include/clap/ext/draft/midi-mappings.h

the midi-mappings extension proposed here?

Bremmers: midi-mappings.h doesn't provide a way to say how important a param is. So a plugin would have to list any mapped params in midi-info.h again.

One could argue these two extensions could/should be rolled into one.

Bremmers: Well, the two extensions aren't really about the same thing: midi-mappings.h is about describing which param/MIDI controller pairs do the same thing. midi-info.h is about describing which MIDI messages work and what the important ones are.

abique:
Hi,
I'm not familiar with the MIDI info, is it part of MIDI 2.0?
Given that we have raw MIDI 2.0 already do we need this? Isn't this already part of the MIDI 2.0 protocol?

Bremmers:
"MIDI info" is just the name of the extension. A host needs this info from the plugin if if wants to support parts of Property Exchange which is part of MIDI-CI (CI = Capability Inquiry). MIDI-CI is part of MIDI 2.0, but it can also be implemented by MIDI 1.0 devices. It's built on sysex messages.

One of the Property Exchange resources is 'ChCtrlList', which is a JSON array describing all MIDI controls the receiver responds to. Here's a super-short one, as an example:

[
 { 
 "title": "Volume",
 "ctrlType": "cc",
 "ctrlIndex": [7],
 },
 { 
 "title": "Modulation Wheel",
 "ctrlType": "cc",
 "ctrlIndex": [1],
 },
] 

A sender (keyboard) receives this info from a receiver (DAW, synth etc.), so it can configure itself to work with the receiver. That's the thing MIDI-CI brings to the table: the sender knows things about the receiver, which is something new for both MIDI and CLAP.

abique:
So the benefit of using this extension versus sysex is that you can query it on the main thread?

include/clap/ext/draft/midi-info.h
```diff
@@ -0,0 +1,94 @@
...
+   // Fills midiinfo. Returns true on success.
+   // [main-thread]
+   bool(CLAP_ABI *get_info)(const clap_plugin_t *plugin,
+                            int16_t              port_index,
+                            uint16_t             channel,
+                            uint32_t             index,
+                            clap_param_info_t   *midiinfo);
+
+   //if false all channels are the same, so the host doesn't have to scan them all.
+   bool(CLAP_ABI *is_multitimbral)(const clap_plugin_t *plugin,
```
abique, line 79: It is missing thread specification.

baconpaul:
Here's the thing I don't quite understand @Bremmers (and it really is ignorance not challenge)

1. The MIDI 2.0 CI spec is huge - I just went and looked - and this extension has a very small set of API points. Things like hand shakes, authority levels, etc... seem very mechanical, but things like Protocol exchange seem on the same level as the sub-set of the API in this extension.
2. As Alex said we have midi 2.0 messages; what is the difference between this and them? Can we just generate this as midi-2.0 messages on the audio thread? What is different about having it as an extension?
3. At an absolute minimum, the extension is misnamed. "midi-info.h" is "information about midi" whereas this is definitely "implementation of the midi 2 property exchange" so at least we should rename it to `midi2-property-exchange.h` and associated. And perhaps have a `midi2-protocol-exchange.h` extension parallel (or perhaps a `midi2-ci.h` which contains both
4. I still think we should segregate out the note expression support into a separate draft extension which is not buried in midi and vst wrapper copies. I don't have time to do that right now but can propose something if you want.

Bremmers:
@baconpaul 
1. The hand shakes etc. are for communication over a wire, where the other end may fail to respond. These problem don't exists at the host/plugin interface. It takes like 10 different sysex messages going back and forth to retrieve the info this extension is about BTW, so a higher level API makes sense I think.

The other 'big thing' besides Property Exchange is Profiles. Protocol negotiation was moved from MIDI-CI to the UMP protocol, I suggest you just ignore that bit.

2. First, this is about meta information about MIDI messages (names etc.). Second, this information goes in opposite direction: from the MIDI receiver (the plugin) to the sender (the DAW, and with that the keyboard).

MIDI-CI is bidirectional. A MIDI-CI enabled piano keyboard which sticks to the old MIDI 1.0 standard would have two DIN cables, the second one is for received data from the DAW or synth it's connected to. But it's still just a piano keyboard that doesn't include sounds. In fact, such a device seems to be coming soon: https://www.gearnews.com/leak-korg-keystage-midi-2-0-controller-with-polyat/

3. While I don't care about the name I think "information about midi" is a good way to think about this. It's not necessarily limited to property exchange. MultitrackStudio, for example, presents the user nice lists of MIDI controllers that actually work with the current instrument, including correct names. This obviously works with the built-in instruments, and now even with MIDI-CI enabled hardware instrument (https://www.kvraudio.com/news/bremmers-audio-design-adds-midi-ci-support-to-multitrackstudio-58816). It doesn't work for plugins because, well, a host knows nothing about a plugin's MIDI implementation. This extension fills that gap. Having to read the manual to see if a synth responds to ccWhatever is so 1983 ;-)

There was a PR for a profiles extension which I accidentally deleted two days ago <blush>. It's still here: https://github.com/Bremmers/clap/blob/Bremmers-midici-profiles/include/clap/ext/draft/midici-profiles.h
It's much more complicated than it seems though, so I'm not in a hurry with that one.

4. I agree with that. I couldn't find that in the wrapper repo. Can you point me to it?

IIRC we explicitly didn't do things like this because it can lead to problems if a plugin supports a certain feature but fails to advertise it correctly, or vice versa. But hosts need to know...

include/clap/ext/draft/midi-info.h
Bremmers: Will fix that!

Bremmers:
> So the benefit of using this extension versus sysex is that you can query it on the main thread?

It also defines a way to get information out of a note input port. The information is going in opposite direction.

baconpaul:
https://github.com/free-audio/clap-wrapper/blob/7cc84d49aebc1d3740f2ab9aa92e01b7555ec712/include/clapwrapper/vst3.h#L106

there's the wrapper code.

your comment #3 - "having to read the manual to decide mid cc whatever is whatever" - makes me think this really is tightly coupled with the midi param mapping somehow.

Bremmers:
> there's the wrapper code.

Hmm, I actually did find that file by myself, but stopped reading two lines before the good bit started. Pfff...

Looks good to me. If it ends up being a standalone extension it would need a host interface with a 'changed' method I think.

> your comment #3 - "having to read the manual to decide mid cc whatever is whatever" - makes me think this really is tightly coupled with the midi param mapping somehow.

I see your point. I wonder why exactly we do have https://github.com/free-audio/clap/blob/main/include/clap/ext/draft/midi-mappings.h in the first place. @abique ?

baconpaul:
Yeah my view, and again remember it is a woefully uninformed one so I'm being careful to admit that, is that we probably want extensions which

1. Allow you to advertise note expression support (with the implicit assumption being that without it you take any)
2. Provide midi property exchanges completely, with a name like midi2-property-exchange. This is a good start for that
3. Provide a way for those properties to also associate with params for daws. midi1-cc-param-mapping is probably the right name for that.
4. It might be that 2 subsumes 3; or 2 conflicts with 3; or 2 and 3 can and should safely co-exist but we should figure that out and document it if we have both; and figure it out and document why we dropped one if not

again: woefully under informed. I have never touched a midi 2 device.

abique:
> I see your point. I wonder why exactly we do have https://github.com/free-audio/clap/blob/main/include/clap/ext/draft/midi-mappings.h in the first place. @abique ?

It is an old thing. I don't see it being useful because it doesn't allow sophisticated mappings.
We can probably throw it away.

PR #353, opened 2023-10-11
clarify latency is defined in samples

defiantnerd:
using the same wording as in tail extension to clarify details.


Bremmers (approved):

abique:
Thanks :+1:

PR #358, opened 2023-11-02
Add flag for enum parameters

messmerd:
See the proposal here: https://github.com/free-audio/clap/discussions/357

This PR adds a `CLAP_PARAM_IS_ENUM` flag to `clap_param_info_flags` to denote an enum parameter.

This flag tells the host when a parameter represents a set of named options mapped to integers, which can help the host decide how to present the parameter to users in its UI.

### DIFF
--- a/include/clap/ext/params.h
+++ b/include/clap/ext/params.h
@@ -195,6 +195,12 @@ enum {
    // A simple example would be a DC Offset, changing it will change the output signal and must be
    // processed.
    CLAP_PARAM_REQUIRES_PROCESS = 1 << 15,
+
+   // The parameter represents a set of named options mapped to integers.
+   // It implies that the parameter is stepped.
+   //
+   // This flag may help the host decide how to present the parameter to users.
+   CLAP_PARAM_IS_ENUM = 1 << 16,
 };
 typedef uint32_t clap_param_info_flags;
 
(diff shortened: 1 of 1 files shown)


defiantnerd:
I think this is really useful PR.

abique:
Maybe it needs a comment to say that the enum isn't allowed to have "holes" between min and max?

defiantnerd:
> Maybe it needs a comment to say that the enum isn't allowed to have "holes" between min and max?

Maybe, although it should be totally clear. You can't have holes in a normal parameter, too.

messmerd:
Maybe I could add this wording?
> _These named options should be available via value_to_text() for each integer value within its [min, max] range._

abique (changes requested):
If the param have `IS_ENUM` then it must have `IS_STEPPED`.

I prefer to say it that way rather than implies.

abique:
Maybe I'd put the comment like this.

```C
// Indicates that this parameter represents an enumerated value.
// If you set this flag, then you must set CLAP_PARAM_IS_STEPPED too.
// All values from min to max must have a non blank value_to_text().
```

abique:
Thank you :+1:

baconpaul:
Oh this is great and I'll use it all the time!

PR #359, opened 2023-11-03
CLAP 1.1.10

abique:


baconpaul (approved):

defiantnerd (approved):
Enum is great, documentation changes are cool.

Bremmers (approved):

include/clap/ext/params.h
```diff
@@ -256,27 +256,32 @@ typedef struct clap_plugin_params {
...
+   // Returns true on success.
    // [main-thread]
    bool(CLAP_ABI *get_value)(const clap_plugin_t *plugin, clap_id param_id, double *out_value);
 
    // Fills out_buffer with a null-terminated UTF-8 string that represents the parameter at the
-   // given 'value' argument. eg: "2.3 kHz". Returns true on success. The host should always use
-   // this to format parameter values before displaying it to the user. [main-thread]
+   // given 'value' argument. eg: "2.3 kHz". The host should always use this to format parameter
+   // values before displaying it to the user.
+   // Returns true on success.
```
messmerd, line 274: For the `value_to_text()` function, I think it would be useful if the host could know when conversion from the double `value` to the string in `out_buffer` is trivial (i.e. 1.0 --> "1.0").

I *could* compare the string `std::to_string(value)` with the string returned from `value_to_text()`, but they might still be different strings due to precision or locale differences in floating point to string conversions. (for example, "1.0000000" and "1.0" compare as false). This makes it difficult and inefficient for hosts to know whether the parameter value to string conversion was trivial.

In LMMS (and any presumably other DAWs that use plain integers or floats when editing parameter automation), this often leads to me having to display two strings to users alongside each other - "1.0" (the parameter's internal value which users use to automate the parameter) and "1.0000" (the supposedly fancy parameter text from `value_to_text()`).

To fix this issue, maybe the return value of `value_to_text()` could be `false` when an error occurred OR the conversion to string would be trivial? This would be a recommendation for plugins to return `false` to let the host do a trivial double-to-string conversion instead of doing that conversion themselves.

include/clap/ext/draft/midi-mappings.h
```diff
@@ -27,6 +27,7 @@ typedef struct clap_plugin_midi_mappings {
    uint32_t(CLAP_ABI *count)(const clap_plugin_t *plugin);
 
    // [main-thread]
+   // Returns true on success and stores the result into mapping.
```
Trinitou, line 30:
```suggestion
   // Returns true on success and stores the result into mapping.
   // [main-thread]
```

include/clap/ext/draft/preset-load.h
```diff
@@ -11,6 +11,7 @@ extern "C" {
 typedef struct clap_plugin_preset_load {
    // Loads a preset in the plugin native preset file format from a location.
    // The preset discovery provider defines the location and load_key to be passed to this function.
+   // Returns true on success.
    //
    // [main-thread]
```
Trinitou, line 16:
```suggestion
   // Returns true on success.
   // [main-thread]
```

include/clap/ext/gui.h
```diff
@@ -204,25 +204,25 @@ typedef struct clap_host_gui {
...
-    * [thread-safe & !floating] */
+   // Request the host to resize the client area to width, height.
+   // Return true if the new size is accepted, false otherwise.
+   // The host doesn't have to call set_size().
+   //
+   // Note: if not called from the main thread, then a return value simply means that the host
+   // acknowledged the request and will process it asynchronously. If the request then can't be
+   // satisfied then the host will call set_size() to revert the operation.
+   //
+   // [thread-safe & !floating]
```
Trinitou, line 215:
```suggestion
   // satisfied then the host will call set_size() to revert the operation.
   // [thread-safe & !floating]
```

include/clap/plugin.h
```diff
@@ -59,6 +59,7 @@ typedef struct clap_plugin {
    // call. The process's sample rate will be constant and process's frame count will included in
    // the [min, max] range, which is bounded by [1, INT32_MAX].
    // Once activated the latency and port configuration must remain constant, until deactivation.
+   // Returns true on success.
    //
    // [main-thread & !active_state]
```
Trinitou, line 64:
```suggestion
   // Returns true on success.
   // [main-thread & !active_state]
```

Trinitou (changes requested):
Just some small format suggestions

include/clap/ext/params.h
```diff
@@ -195,6 +195,11 @@ enum {
    // A simple example would be a DC Offset, changing it will change the output signal and must be
    // processed.
    CLAP_PARAM_REQUIRES_PROCESS = 1 << 15,
+
+   // This parameter represents an enumerated value.
+   // If you set this flag, then you must set CLAP_PARAM_IS_STEPPED too.
+   // All values from min to max must not have a blank value_to_text().
+   CLAP_PARAM_IS_ENUM = 1 << 16,
```
messmerd, line 202:
```suggestion

   // This parameter represents an enumerated value.
   // All values from min to max must not have a blank value_to_text().
   CLAP_PARAM_IS_ENUM = CLAP_PARAM_IS_STEPPED | (1 << 16),
```
Maybe we should express programmatically that enum parameters must also be CLAP_PARAM_IS_STEPPED?

include/clap/ext/params.h
baconpaul: Why do your users edit plain floats? To edit a parameter they should type in a string which the host hands to string to value and let the parameter handle it no?

include/clap/ext/params.h
Trinitou: I think that wouldn't prevent programmers from manually assigning `CLAP_PARAM_IS_ENUM & ~(CLAP_PARAM_IS_STEPPED)`. So it still is important to clearly document the correct behaviour vs. only doing it programmatically. Also would be a breaking change as existing checks of `flags & CLAP_PARAM_IS_ENUM` would then also return true if only `CLAP_PARAM_IS_STEPPED` is present in `flags`

abique: No this is dodgy we've tried that in the past.
The problem is that `(flags & CLAP_PARAM_IS_ENUM) != 0` if flags is only stepped.

defiantnerd: I am fine with `CLAP_PARAM_IS_STEPPED` is implied, when `CLAP_PARAM_IS_ENUM` is set. Stepped does not mean Enum, but Enum just implies it. And the helpers could even check for `CLAP_PARAM_IS_STEPPED` being set when `CLAP_PARAM_IS_ENUM` is set, too. 

We should set one of the flags having more than one bit, this will add confusion and does not match a good code style.

include/clap/ext/params.h
messmerd: LMMS uses plain float or integer values in the automation editor and parameter editing dialog boxes, not custom strings like those returned from `value_to_text()`. The parameter/automation system could potentially be reworked in the future, but would require a lot of work just to better support CLAP.

Until that happens, LMMS and any other hosts that do not (yet) support CLAP's recommended practice of hiding internal parameter values from users and only exposing strings from `value_to_text()` will have an unnecessarily rough time which I believe could be alleviated if CLAP provided better guidance to plugin devs on how to implement `value_to_text()`.

That is, use a return value of `false` to indicate that a particular parameter has no special way of presenting itself and can just be trivially converted to a string. This is similar to how leaving `value_to_text()` unimplemented indicates that none of the parameters have a special way of presenting themselves.

I understand if this is not a change you'd like to make, but to me it is better to specify the meaning of the `false` return value AND make it useful for certain hosts than it is to leave it unspecified and allow diverging implementations in plugins.

baconpaul: yeah i view 'false' as in 'i am unable to convert this value to a string'. for instance if you send an out of range value or haven't implemented the clap api. I don't think we should overload it to mean "I did nothing"; but rather it means "don't trust the value of the text buffer as being correctly populated". We could definitely add this to do the doc if people agree, though.

I understand and am sympathetic to the work in LMMS to rework to use strings through the editor. But I'm not that thrilled about changing the base API for that (although others may disagree). Having the param value <> string be bidirectional and both as a requirement of the plugin developer and a guide for the host author seems good.

Perhaps one solution would be to offer an LMMS specific extension where you could advertise this fact? 

But also: taking a quick look at surge and the conduit plugins, I think none of our few hundred params would return true - maybe the 8 macros would but I think even those we convert to % * 100 at the edge. I'm not sure how other plugin authors would score.

baconpaul: (Also I'm very excited for LMMS to get CLAP support; many of our users ask for quality electronic open source DAWs which can support Surge and since we don't ship a VST2 etc... if I can help separately from this issue please always do tag me in)

messmerd: Just to be clear, this isn't a roadblock to implementing CLAP support in LMMS - it's just (in my opinion) a "nice to have" enhancement. If the `value_to_text()` function remains the same as it is, that's fine with me. I just figured it would be worth a shot suggesting a way to change it.

I also appreciate your offer to help regarding Surge support in LMMS. In the next few weeks, I'll be trying to get Surge to an operational state in LMMS's CLAP host, so I'll let you know if any issues come up that I don't know how to handle.

robbert-vdh (approved):
LGTM.

abique:
Thank you everyone!

PR #360, opened 2023-11-03
Clarify the meaning of boolean return values

messmerd:
Many functions that return `bool` do not specify what that value means, so I marked those functions with "Returns true on success" or similar wording.

Along the way I also fixed a few typos and style issues.

### DIFF
--- a/include/clap/ext/draft/triggers.h
+++ b/include/clap/ext/draft/triggers.h
@@ -88,7 +88,6 @@ typedef struct clap_plugin_triggers {
    uint32_t(CLAP_ABI *count)(const clap_plugin_t *plugin);
 
    // Copies the trigger's info to trigger_info and returns true on success.
-   // Returns true on success.
    // [main-thread]
    bool(CLAP_ABI *get_info)(const clap_plugin_t *plugin,
                             uint32_t             index,
--- a/include/clap/ext/note-ports.h
+++ b/include/clap/ext/note-ports.h
@@ -40,7 +40,7 @@ typedef struct clap_note_port_info {
 
 // The note ports scan has to be done while the plugin is deactivated.
 typedef struct clap_plugin_note_ports {
-   // Number of ports, for either input or output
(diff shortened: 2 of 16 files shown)


include/clap/ext/timer-support.h
```diff
@@ -16,10 +16,11 @@ typedef struct clap_plugin_timer_support {
 typedef struct clap_host_timer_support {
    // Registers a periodic timer.
    // The host may adjust the period if it is under a certain threshold.
-   // 30 Hz should be allowed.
+   // 30 Hz should be allowed. Returns true on success.
```
abique, line 19: Maybe line break here?

include/clap/ext/note-ports.h
```diff
@@ -40,11 +40,12 @@ typedef struct clap_note_port_info {
 
 // The note ports scan has to be done while the plugin is deactivated.
 typedef struct clap_plugin_note_ports {
-   // number of ports, for either input or output
+   // Number of ports, for either input or output
```
abique, line 43: If you start with capital, then end with `.` maybe?

include/clap/ext/draft/triggers.h
```diff
@@ -88,7 +88,6 @@ typedef struct clap_plugin_triggers {
    uint32_t(CLAP_ABI *count)(const clap_plugin_t *plugin);
 
    // Copies the trigger's info to trigger_info and returns true on success.
-   // Returns true on success.
```
abique, line 91: Why did you remove that one?

abique:
Thank very much you for this change, please address the comments and it should be good :+1:

include/clap/ext/draft/triggers.h
messmerd: The line above it already mentioned "returns true on success"

abique: Ah correct!

PR #364, opened 2023-11-21
Unlikely

Trinitou:
the `unlikely` attribute is used in the `clap-host` repo but is C++20 or higher (so not C++17 friendly)

I guess it's good to define it here(?)
Or should it rather be in the `clap-host` repository?


abique:
I'd rather avoid it.
Because it doesn't have the same syntax in C and C++. I think in C you have to do `__unlikely__((expr))`.
Also clap itself almost ship no code so it doesn't need it.

clap-helpers could have it.

PR #366, opened 2023-11-28
Clarify the allowed usages of clap_plugin_entry's init() and deinit()

prokopyl:
This PR is a followup to the discussion in #365, and clarifies the following points in the specification of `clap_plugin_entry`:

* Assert that a CLAP DSO *can* be re-`init`ialized after it has been `deinit`ialized.
  
   The specification currently states that `init` "can only be called once", and that `deinit` prevents "[any] more calls to into the DSO". These can be interpreted as "`init` can only be called once ever" and "nothing in the DSO cannot be used ever again after `deinit`", which would be an odd requirement, and indeed seems not to be the intent of the spec according to the discussion in #365.

   This PR changes those statements to clearly indicate that re-initialization is allowed, while still forbidding to call `init` and `deinit` again if the DSO has already been initialized/de-initialized respectively (which I believe is the intent of the original wording).
* Replace the assertion that confusingly marks `init` and `deinit` as "thread-safe", with a more specific explanation of the threading rules for these two functions.
  
  Using the "thread-safe" term is confusing here because it is usually interpreted as "can be safely called simultaneously from multiple threads", but consecutive calls to these functions are explicitly forbidden, let alone simultaneous ones.

   This is replaced with a more detailed definition, which explicitly forbids simultaneous calls, while also explaining that the host may call `init` and `deinit` from any (and different) threads.

   This PR also makes it explicit that those threading requirements do not apply to `get_factory`, which is thread-safe, and now uses the standard `[thread-safe]` thread specification.

I also took the liberty to fix a couple of unrelated typos in those specifications. :slightly_smiling_face:

### DIFF
--- a/include/clap/entry.h
+++ b/include/clap/entry.h
@@ -30,35 +30,63 @@ extern "C" {
 //
 // Each directory should be recursively searched for files and/or bundles as appropriate in your OS
 // ending with the extension `.clap`.
-//
-// Every method must be thread-safe.
 typedef struct clap_plugin_entry {
    clap_version_t clap_version; // initialized to CLAP_VERSION
 
-   // This function must be called first, and can only be called once.
+   // Initializes the DSO.
+   //
+   // This function must be called first, before any-other CLAP-related function or symbol from this
+   // DSO.
+   //
+   // It also must only be called once, until a later call to deinit() is made, after which init()
+   // can be called once more to re-initialize the DSO.
+   // This enables hosts to e.g. quickly load and unload a DSO for scanning its plugins, and then
+   // load it again later to actually use the plugins if needed.
    //
    // It should be as fast as possible, in order to perform a very quick scan of the plugin
    // descriptors.
    //
-   // It is forbidden to display graphical user interface in this call.
-   // It is forbidden to perform user interaction in this call.
+   // It is forbidden to display graphical user interfaces in this call.
+   // It is forbidden to perform any user interaction in this call.
    //
    // If the initialization depends upon expensive computation, maybe try to do them ahead of time
    // and cache the result.
    //
-   // If init() returns false, then the host must not call deinit() nor any other clap
-   // related symbols from the DSO.
+   // Returns true on success. If init() returns false, then the DSO must be considered
+   // uninitialized, and the host must not call deinit() nor any other CLAP-related symbols from the
+   // DSO.
    //
    // plugin_path is the path to the DSO (Linux, Windows), or the bundle (macOS).
+   //
+   // This function may be called on any thread, including a different one from the one a later call
+   // to deinit() (or a later init()) can be made.
+   // However, it is forbidden to call this function simultaneously from multiple threads.
+   // It is also forbidden to call it simultaneously with *any* other CLAP-related symbols from the
+   // DSO, including (but not limited to) deinit().
    bool(CLAP_ABI *init)(const char *plugin_path);
 
-   // No more calls into the DSO must be made after calling deinit().
+   // De-initializes the DSO, freeing any resources allocated or initialized by init().
+   //
+   // After this function is called, no more calls into the DSO must be made, except calling init()
+   // again to re-initialize the DSO.
+   // This means that after deinit() is called, the DSO can be considered to be in the same state
+   // as if init() was never called at all yet, enabling it to be re-initialized as needed.
+   //
+   // Just like init(), this function may be called on any thread, including a different one from
+   // the one init() was called from, or from the one a later init() call can be made.
+   // However, it is forbidden to call this function simultaneously from multiple threads.
+   // It is also forbidden to call it simultaneously with *any* other CLAP-related symbols from the
+   // DSO, including (but not limited to) deinit().
    void(CLAP_ABI *deinit)(void);
 
    // Get the pointer to a factory. See factory/plugin-factory.h for an example.
    //
    // Returns null if the factory is not provided.
    // The returned pointer must *not* be freed by the caller.
+   //
+   // Unlike init() and deinit(), this function can be called simultaneously by multiple threads.
+   //
+   // [thread-safe]
    const void *(CLAP_ABI *get_factory)(const char *factory_id);
 } clap_plugin_entry_t;
 
--- a/include/clap/entry.h
+++ b/include/clap/entry.h
@@ -30,6 +30,43 @@ extern "C" {
 //
 // Each directory should be recursively searched for files and/or bundles as appropriate in your OS
 // ending with the extension `.clap`.
+//
+// As detailed in the respective functions' documentations below, the host is forbidden from calling
+// either init() or deinit() multiple times, without calling the other first. This is to enforce
+// a "normal" initialization pattern, where one can't construct or destroy an object multiple times.
+//
+// However, there are a few cases where the host may not be able to reliably prevent this from
+// happening, such as when multiple, separate hosts co-exist within the same address space (e.g.
+// Meta-Plugins, plugins that can host other plugins inside them). It may also happen when a VST
+// compatibility layer is used to wrap a CLAP plugin, for instance.
+//
+// In those cases, the init() and deinit() pair of functions can happen to be called multiple times,
+// either in a row or even simultaneously, from different threads, by multiple non-synchronized
+// hosts, or by a single host that is presented multiple interfaces to a same DSO. This can happen
+// even though this is forbidden behavior from the host by this specification, and despite a
+// compliant host's best efforts.
+//
+// Despite those issues, hosts *must* ensure that each successful call to init() is only followed by
+// a single call to deinit() at most. As long as a single successful init() call did not have its
+// matching deinit() called, the DSO *must* remain in a fully operational state, as if deinit() had
+// not yet been called.
+//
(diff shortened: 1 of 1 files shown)


abique:
Hi,

There is one issue with this specification, which wasn't solved before anyway but it should be addressed if we change this file.

Take the following scenario:
- Host loads plugins A and B
- Plugin A itself hosts plugins and loads B

Here A has no knowledge that the host had loaded B.

Here's one way for the plugin to defensively implement init:
```C++
static std::mutex g_plugin_init_mutex;
static int g_init_count = 0;

static bool do_init(const char *plugin_path)
{
   /* actual init function */
}

static bool init_guard(const char *plugin_path)
{
   std::lock_guard<std::mutex> guard(g_plugin_init_mutex);

   ++g_plugin_init_count;
   if (g_plugin_init_count > 1)
      return true;

   if (do_init(plugin_path))
      return true;

  g_plugin_init_count = 0;
  return false;
}

static void do_deinit()
{
   /* actual deinit function */
}

static void deinit_guard()
{
   std::lock_guard<std::mutex> guard(g_plugin_init_mutex);

   if (g_plugin_init_count == 0)
     return;

   --g_plugin_init_count;
   if (g_plugin_init_count == 0)
      do_deinit();
}
```

I think the correct spec would be:
- thread safe
- init counter
- if `init()` did return true, then there must be a corresponding call to `deinit()`

abique:
> However, I don't think we can allow multiple subsequent calls to `init` now (by forcing plugins to use an init counter as you suggested, or any other defense mechanism), as this would be a breaking change to the spec. In this case it could potentially lead to crashes if a host wanted to leverage this new property on a plugin that hasn't been updated.

We may allow it or not, it may still happen.

> Moreover, I don't think this solution would help much in the scenario you described, in practice. There are plugins out there that aren't that defensive in their `init` implementations (and correctly so, since they relied on the spec that allows them not to be), which means that "sub-hosts", like Plugin A in your example, cannot rely on a plugin-side implementation anyway.

It is fine to not be defensive, if you enjoy customer support :)

> As of now, the only solution for "sub-hosts" like this (that I can think of, at least 🙂), is to spawn their own child process (like Yabridge does) to have their own address space for their plugins.

Indeed, but that's quite a requirement versus make your `init()` function robust.

> Alternatively, since the root of the issue here is that "A has no knowledge that the host had (un)loaded B", we could perhaps come up with an extension instead, that would allow for the parent host and the sub-host (A) to synchronize their init counters for a specific plugin's DSO, so that a sub-host doesn't double-load (or unload) a plugin that the parent host is using, and vice-versa. I have some ideas, but of course that would be outside the scope of this PR. 🙂

Still, it doesn't work.

You have a vst2 host, hosting plugin A (vst2) and B (vst2), and both A and B open a plugin C (CLAP).

You may also have a plugin that is symlinked into the vst2 and clap plugin folder, then `dlopen()` will open the same file, so you may end up with the dual `init()` anyway.

I think the only solution, even today is to have a defensive implementation of the init/deinit thing.

baconpaul:
So all of my `init` methods are `return true` but I agree with Alex, there is no way for a host to guarantee the spec requirement that `init` is only called once per dlopen and as such if you do something non-trivial you have to defend. In the very real case, for instance, of running the clap and the vst3 wrapped clap in the same process space, you get a double init and (in some cases) double dlopen.

I wish we had realized this earlier of course, but I think we have to at least share that defacto reality in the spec. inits and de-inits match and can be called from any thread is, I think, the current state of affairs of what actually happens to a plugin, independent of the spec.

Whether we update the documentation to say that is an interesting question about how we manage changes in the spec.

prokopyl:
> > However, I don't think we can allow multiple subsequent calls to `init` now (by forcing plugins to use an init counter as you suggested, or any other defense mechanism), as this would be a breaking change to the spec. In this case it could potentially lead to crashes if a host wanted to leverage this new property on a plugin that hasn't been updated.
> 
> We may allow it or not, it may still happen.

That's true, my concern is that it may happen *more* if we allow it, since changing it that way allows hosts to be *less* defensive on their end. As of now, hosts that call `init` multiple times themselves (before a `deinit`) are non-compliant to the spec (actually [been there, done that recently](https://github.com/prokopyl/clack/pull/16) :sweat_smile:), which helps reduce the likelihood of this problem occurring.

> > Moreover, I don't think this solution would help much in the scenario you described, in practice. There are plugins out there that aren't that defensive in their `init` implementations (and correctly so, since they relied on the spec that allows them not to be), which means that "sub-hosts", like Plugin A in your example, cannot rely on a plugin-side implementation anyway.
> 
> It is fine to not be defensive, if you enjoy customer support :)

True haha. Here I was thinking about what happens if Plugin B isn't updated anymore and can't catch up to the new spec. You won't get more customer support issues if you don't support customers at all in the first place. :wink: 

(Plugin A will likely get a bunch of compat issues for plugin B though.)

> > As of now, the only solution for "sub-hosts" like this (that I can think of, at least 🙂), is to spawn their own child process (like Yabridge does) to have their own address space for their plugins.
> 
> Indeed, but that's quite a requirement versus make your `init()` function robust.

I don't see how making your *own* `init()` function would help here? As I understand it, *every* plugin out there would need to be patched to make their `init()` function robust to properly support this use case. If I was Plugin A's author, I'd much rather be more defensive and implement the child process solution, rather than trying to hunt down all of the plugins that aren't defensive enough out there.

> > Alternatively, since the root of the issue here is that "A has no knowledge that the host had (un)loaded B", we could perhaps come up with an extension instead, that would allow for the parent host and the sub-host (A) to synchronize their init counters for a specific plugin's DSO, so that a sub-host doesn't double-load (or unload) a plugin that the parent host is using, and vice-versa. I have some ideas, but of course that would be outside the scope of this PR. 🙂
> 
> Still, it doesn't work.
> 
> You have a vst2 host, hosting plugin A (vst2) and B (vst2), and both A and B open a plugin C (CLAP).

Ah, right, I didn't think about that. I use so little of these MetaPlugin-like wrappers, I didn't even think one could have *two* of them! :sweat_smile: 

I also forgot about the VST -> CLAP wrappers, yeah I don't see how hosts could possibly defend against double-init in those cases.

> I think the only solution, even today is to have a defensive implementation of the init/deinit thing.

Yeah, I agree. We could heavily suggest that plugins should be defensive in their init/deinit implementations, while still forbidding hosts from do calling `init` twice directly like the spec does now. I can make that addition in this PR. :slightly_smiling_face:

prokopyl:
> We could make the spec say that any plugin build with clap >= 1.11.0 should implement the counter behavior.

Well I should check for new replies before clicking Send lol. Disregard my previous message, I'll just do that. :smile:

baconpaul:
I think we should make it say

"Prior to clap 1.11 our documentation stated that init was called once, and deinit once. Unfortunately in some circumstances it is impossible for a host to realize this, thus with clap 1.11 we are simply making the statement that init calls and deinit calls must match. This means the de-facto practice of protecting your init with a locked counter is now the to-spec practice also."

or some such. See what I mean? Don't make it sound like you need "if version < 1.11 then don't lock and count else lock and count" in the docs

prokopyl:
Yes I see what you mean. I think the best way to do this would be to assert that plugins *should* implement the locked counter defense strategy (since the issue affects all CLAP versions anyway), and that this becomes a hard requirement as of CLAP v1.11.

While adding a hard requirement is still a breaking change technically, it doesn't break *more* compared to the way it is done now, so that works for me. ^^

baconpaul:
yup that's the spirit of what i was getting at too. Agree.

prokopyl:
I've added that in my latest commit. It's a quite a lot, but I think I've got all of the details and possible side-effects of the issue covered now. :slightly_smiling_face:

abique:
Hi,

Thank you very much @prokopyl .

I went through it, I think the text is long for my taste, but the history told us that long is better than too short.
@baconpaul have you gone through it too? What do you think?

I think it looks good  for the merge, we'll still have the possibility to ajust it in the final 1.11.0 review.

abique:
My style would be more condensed:
1. everything is thread safe
2. init counter
3. note for plugins with clap <1.11.0 which didn't have the init counter requirement
4. Rationale explanation of the design

My issue with a documentation that is too long is that devs don't read it, and when they reach the bottom of it, they forgot what was in the top, so it requires multiple reads.

With that plan you'd get the essential information immediately, and then it is mostly the rationale that takes more line, but the brain already has the big picture when reading it.

abique:
I've implemented the entry init counter in the plugin template.

baconpaul:
Right. I agree 

so let’s have a tldr and move that text to later in the file

thentldr is

“init and deinit in most cases are called once, in a matched pair, when the dso is loaded / unloaded. In some rare situations it may be called multiple times in a process, so the functions must be defensive, mutex locking and counting calls if undertaking non trivial non idempotent actions.”

abique:
Sorry, I forgot about this PR.
I review it now.

abique:
I've merged it but I wish @prokopyl did the documentation update that was suggested in the discussion.

prokopyl:
My apologies for the delay. I know it's quite a bit late now that 1.2.0 is out, but I can still make another PR to incorporate the suggestions if you wish.

PR #367, opened 2023-12-05
Adjust active and processing state documentation

Trinitou:


include/clap/ext/audio-ports.h
```diff
@@ -65,15 +65,14 @@ typedef struct clap_audio_port_info {
    clap_id in_place_pair;
 } clap_audio_port_info_t;
 
-// The audio ports scan has to be done while the plugin is deactivated.
 typedef struct clap_plugin_audio_ports {
    // Number of ports, for either input or output
-   // [main-thread]
+   // [main-thread & !active]
```
abique, line 70: The ports layout has to be known before the plugin is activated, but then why would you forbid reading the ports?

include/clap/ext/thread-pool.h
```diff
@@ -57,7 +57,7 @@ typedef struct clap_host_thread_pool {
    // Returns true if the host did execute all the tasks, false if it rejected the request.
    // The host should check that the plugin is within the process call, and if not, reject the exec
    // request.
-   // [audio-thread]
+   // [audio-thread & active & processing]
    bool(CLAP_ABI *request_exec)(const clap_host_t *host, uint32_t num_tasks);
```
abique, line 61: I think audio-thread already implies active and processing, but it doesn't hurt to repeat.

abique:
Thanks!

PR #369, opened 2023-12-18
Correct referenced function name in comment

cubicvoid:
Correct a comment in `params.h` that misreferences `clap_plugin_params.get_value()` as `clap_plugin_params.value()`


abique:
Thank you! :+1:

PR #371, opened 2023-12-20
CLAP 1.2.0

abique:


ChangeLog.md
```diff
@@ -1,3 +1,44 @@
...
+* `CLAP_EXT_CONFIGURABLE_AUDIO_PORTS`
+* `CLAP_EXT_CONTEXT_MENU`
+* `CLAP_EXT_PARAM_INDICATION`
+* `CLAP_EXT_PRESET_LOAD`
+* `CLAP_EXT_REMOTE_CONTROLS`
+* `CLAP_EXT_STATE_CONTEXT`
+* `CLAP_EXT_SURROUND`
+* `CLAP_EXT_TRACK_INFO`
+
+Note: we kept the last draft extension ID in order to not break plugin already using it.
```
ritschwumm, line 15: typo: "plugin" should be "plugins"

ChangeLog.md
```diff
@@ -1,3 +1,44 @@
...
+
+## Stabilize factory
+* `CLAP_PRESET_DISCOVERY_FACTORY_ID`
+
+Note: we kept the last draft factory ID in order to not break plugin already using it.
```
ritschwumm, line 20: plugins

ChangeLog.md
```diff
@@ -1,3 +1,44 @@
...
+
+## Documentation
+
+* [events.h](include/clap/events.h): Clarify how "Port Channel Key NoteID" matching works
+* [plugin.h](include/clap/plugin.h): Style cleanup
+* [params.h](include/clap/ext/params.h): Fix incorrect function name reference
+* [latency.h](include/clap/ext/latency.h): Precise that the latency can only be fetched when the plugin is activated
```
ritschwumm, line 27: what does this mean? "Clarify that"  or "Be precise" maybe?

ChangeLog.md
```diff
@@ -1,3 +1,44 @@
...
+
+## Conventions
+
+* [extension-id](conventions/extension-id.md): introduce some rules about extension ID naming.
+
+## Plugin Template
+
+* [plugin-template.c](src/plugin-template.c): Add exemple code for safer clap entry initialization
```
ritschwumm, line 35: typo: "exemple" should be "example"

conventions/extension-id.md
```diff
@@ -0,0 +1,30 @@
...
+Where:
+- `$REVERSE_URI` would be something like `com.bitwig`.
+
+## Draft
+
+An extension is considered a draft extension if it is in the [draft](../include/clap/ext/draft/) folder.
+
+When the extension is migrating from draft to stable, its extension ID must not change.
+
+All extensions must go first though a draft phase.
```
ritschwumm, line 20: is the grammar correct here? i it should be "All extensions must go though the draft phase first."

abique: Thanks, I've applied your suggestions.

include/clap/ext/latency.h
```diff
@@ -11,7 +11,7 @@ extern "C" {
 // The audio ports scan has to be done while the plugin is deactivated.
```
Bremmers, line 11: This comment should be removed I think?

abique: Oh...

conventions/extension-id.md
```diff
@@ -0,0 +1,30 @@
...
+
+An extension is considered a draft extension if it is in the [draft](../include/clap/ext/draft/) folder.
+
+When the extension is migrating from draft to stable, its extension ID must not change.
+
+All extensions must go though the draft phase first.
+
+## For factory ID
+
+Everything about the extension id simetrically applies to factory id.
```
messmerd, line 24:
```suggestion
Everything about the extension id symmetrically applies to factory id.
```
typo

include/clap/factory/draft/plugin-state-converter.h
```diff
@@ -0,0 +1,72 @@
...
+   // [thread-safe]
+   bool (*convert_plain_value)(const struct clap_plugin_state_converter *converter,
+                               clap_id                                   src_param_id,
+                               double                                    src_plain_value,
+                               clap_id                                  *dst_param_id,
+                               double                                   *dst_plain_value);
+} clap_plugin_state_converter_t;
+
+// Factory identifier
+static CLAP_CONSTEXPR const char CLAP_CLAP_CONVERTER_FACTORY_ID[] =
```
Bremmers, line 55: CLAP_CLAP_ should probably be CLAP_PLUGIN_STATE_ I think?

conventions/extension-id.md
```diff
@@ -0,0 +1,30 @@
...
+
+## For factory ID
+
+Everything about the extension id symmetrically applies to factory id.
+
+## History
+
+Before this document was written, existing extension didn't honor those rules.
+Some stable extension include the string `draft` in their ID.
+We decided to keep those in order to maintain the binary compatibility.
```
messmerd, line 30:
```suggestion
Before this document was written, existing extensions didn't honor these rules.
Some stable extensions include the string `draft` in their ID.
We decided to keep those in order to maintain binary compatibility.
```
wording improvement

include/clap/factory/draft/plugin-state-converter.h
abique: Yes!

include/clap/entry.h
```diff
@@ -31,34 +31,100 @@ extern "C" {
...
+   // De-initializes the DSO, freeing any resources allocated or initialized by init().
+   //
+   // After this function is called, no more calls into the DSO must be made, except calling init()
+   // again to re-initialize the DSO.
+   // This means that after deinit() is called, the DSO can be considered to be in the same state
+   // as if init() was never called at all yet, enabling it to be re-initialized as needed.
+   //
+   // As stated above, even though hosts are forbidden to do so directly, multiple calls before any
+   // new init() call may still happen. Implementations *should* take this into account, and *must*
+   // do so as of CLAP 1.11.
```
messmerd, line 111:
```suggestion
   // do so as of CLAP 1.2.0.
```

messmerd (commented):
There are some "CLAP 1.1.11" references that could be changed to "CLAP 1.2.0".

Bremmers (approved):

abique:
When going through the macros, we have:
```C
#if defined(CLAP_CPLUSPLUS) && CLAP_CPLUSPLUS >= 201103L
#   define CLAP_HAS_CXX11
#   define CLAP_CONSTEXPR constexpr
#else
#   define CLAP_CONSTEXPR
#endif
```

If I'm correct, in C++ constexpr implies const, so maybe in the else branch `#define CLAP_CONSTEXPR const` would be better?

Then if we do the change above, second question:
```C
static CLAP_CONSTEXPR const char CLAP_EXT_TRACK_INFO[] = "clap.track-info.draft/1";
// OR
static CLAP_CONSTEXPR char CLAP_EXT_TRACK_INFO[] = "clap.track-info.draft/1";
```

@Bremmers @robbert-vdh would that create problems for making rust/delphi/... bindings to CLAP?

Thank you for your advice.

abique:
I'd like to re-discuss the draft transition for 1.2.0.

Breaking the end-user experience isn't an option to me because it will leave a massive negative impression of CLAP. Users will fail to understand why it broke and why we'd make that choice. There will be traces in the forums that will never go away. Overall, it'll have a negative impact on the project.

However, in the set of extensions that we stabilize in 1.2.0, the audio ports related one aren't widely used and are OK to break I suppose, then  for the others, breaking them won't affect the projects playback at least in Bitwig. They are more convenience extensions, like preset-discovery, preset-load, param-indication, remote-controls, ...

I see two possible path:
1. keep the draft in the extension id
   - This will look ugly (in the sources) forever
   - No extra pain for the devs, everything stays working
2. remove the draft in the extension id, introduce a compat ext id eg: `CLAP_EXT_CONTEXT_MENU_COMPAT`, ask the plugin and hosts to check both the ext id and the compat ext id if the compatibility matters to them, remove the compat ext id at some point in the future
   - kind of ugly to have to check for 2 extensions ids, but doable
   - there will be cases where it "doesn't work" because of old and new product combination
   - transition time needs to be 2 years I think
   - code will look good in 2026, if we ever decide to remove the compat ext id, which may decide against in the futur and it'll turn out to stay forever...

What do you think?

Bremmers:
> @Bremmers @robbert-vdh would that create problems for making rust/delphi/... bindings to CLAP?

Would be no problem for the Delphi translation.

messmerd:
If we go with (2), which I'm in favor of, it should not be very difficult to coordinate with the developers of [the handful of known CLAP hosts](https://clapdb.tech/category/hostsdaws) to alert them to this change ahead of time. We could have support for both the new and compatibility IDs in hosts before CLAP 1.2.0 is even released.

I'd even be willing to open PRs in host repos to add this support myself.

If all major hosts support both the old and new IDs before 1.2.0 is released, plugin developers should not have to worry about or even know about this change at all, so I think any `COMPAT` constants should be directed primarily towards host developers who want to support pre-CLAP 1.2.0 plugins with any of the affected extensions, not at plugin developers. I could even see this being a footnote in a comment under the new IDs rather than a `COMPAT` constant that we'd need to remove later.

The only issue I can see is that the affected extensions in new (>=1.2.0) plugins will not work in old (<1.2.0) hosts.

However, this does not really bother me because:

- Most CLAP plugins I've seen currently use older CLAP versions such as 1.1.1 and are not actively upgrading to the newest CLAP version as soon as it releases. This gives hosts more time to catch up.
- Plugin frameworks such as CLAP JUCE extensions and others could delay updating to 1.2.0 if they really want to ease the transition for plugins that rely on them.

baconpaul:
I'm fine with option 1 for the ones we promote to 1.2.0 as long as the ones left in draft in 1.2.0 (1) remove draft from their name and (2) don't auto-include the then-still drafts in clap.h.

I think the draft in the name is less embarrassing than breaking for no reason, I agree. I just don't want to have to make this choice again at 1.3, 1.4, 1.5 etc....

abique:
> I'm fine with option 1 for the ones we promote to 1.2.0 as long as the ones left in draft in 1.2.0 (1) remove draft from their name and (2) don't auto-include the then-still drafts in clap.h.
> 
> I think the draft in the name is less embarrassing than breaking for no reason, I agree. I just don't want to have to make this choice again at 1.3, 1.4, 1.5 etc....

With Bitwig, the break will happen only if:
- bitwig <= 5.0
- plugin uses clap >= 1.2.0 without clap-helpers

The code to work with both ext id would be added to bitwig and clap-helpers.

baconpaul:
So that’s the camp 2 compat Id idea?

I’m fine with that too. 

Like I said my biggest concern is we don’t have to do this again at 1.3 so things that stay draft in 1.2 get draft removed etc

urshimself:
I'm good with 2, too. Most important thing is it'll eventually be resolved, at the end it's a clean resolution, it won't happen again.

That said, we have experienced worse compatibility breaks in all other formats we support, often without even being notified about it. Mostly due to SDK changes or change of interpretation of specs. Sometimes we learn through users.

Therefore the only wish I have is that the changes are communicated transparently, like this is what we did, this is why we did it and that is why it's gonna be resolved forever.

baconpaul:
Yeah it takes me a morning to do a clap point release with a new clap helpers for surge too if helpers does the both names trick

abique:
I've prepared a PR with what changing the stabilized extension ID would look like: https://github.com/free-audio/clap/pull/379

include/clap/ext/ambisonic.h
```diff
@@ -1,9 +1,11 @@
 #pragma once
 
-#include "../../plugin.h"
+#include "../plugin.h"
 
 // This extension can be used to specify the channel mapping used by the plugin.
-
+//
+// This extension ID contains `draft', yet it is stable.
+// See conventions/extension-id.md for more info.
```
baconpaul, line 8: Curious why others got compat ids and this one didn't?

include/clap/ext/state.h
```diff
@@ -10,6 +10,10 @@
 /// values and non-parameter state. This is used to persist a plugin's state
 /// between project reloads, when duplicating and copying plugin instances, and
 /// for host-side preset management.
+///
+/// If you need to know if the save/load operation is meant for duplicating a plugin
+/// instance, for saving/loading a plugin preset or while saving/loading the project
+/// then have a look at clap_plugin_state_context_t.
```
baconpaul, line 16: "then have a look at" may be better stated as "consider implementing state-context in addition to state"

include/clap/plugin-id.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+// Pair of plugin ABI and plugin identifier.
+typedef struct clap_plugin_id {
+   // The plugin ABI name, in lowercase.
+   // eg: "clap"
```
baconpaul, line 6: I think this needs more expansion on what you mean by "ABI Name"; or alternately a set of constexpr char* with examples.

baconpaul (approved):
A few comments, only one of them material. I'm happy to merge as is but if we are still making changes, expanding the comments - especially on `clap_plugin_id::api` - seems smart

include/clap/ext/ambisonic.h
abique: This is an overlook!

baconpaul: Oh cool! Well that's why we do reviews I guess :)

include/clap/plugin-id.h
abique: Do you think it is OK to give VST3 and VST2 examples in the documentation?

include/clap/universal-plugin-id.h
```diff
@@ -0,0 +1,26 @@
...
+   // CLAP: use the plugin id
+   //   eg: "com.u-he.diva"
+   //
+   // AU: format the string like "type:subt:manu"
+   //   eg: "aumu:SgXT:VmbA"
+   //
+   // VST2: print the id as a signed 32-bits integer
+   //   eg: "-4382976"
+   //
+   // VST3: print the id as a standard UUID
```
Bremmers, line 23: Perhaps add "(lower case)" ?

defiantnerd (approved):

include/clap/factory/preset-discovery.h
```diff
@@ -93,17 +99,6 @@ typedef uint64_t clap_timestamp_t;
 // Value for unknown timestamp.
 static const clap_timestamp_t CLAP_TIMESTAMP_UNKNOWN = 0;
```
Trinitou, line 100: PR for renaming `clap_timestamp_t` and moving it to a separate header: #385 
(also CLAP_CONSTEPR was missing for `CLAP_TIMESTAMP_UNKNOWN`

include/clap/clap.h
```diff
@@ -28,44 +28,37 @@
...
 
 #include "factory/plugin-factory.h"
-#include "factory/draft/plugin-invalidation.h"
-#include "factory/draft/preset-discovery.h"
+#include "factory/preset-discovery.h"
 
 #include "plugin.h"
 #include "plugin-features.h"
 #include "host.h"
+#include "universal-plugin-id.h"
```
Trinitou, line 36:
```suggestion
```
is it really necessary to include it here? I mean, it's already included via `#include "factory/preset-discovery.h"`.  Also e.g. `id.h` is also not included here

include/clap/clap.h
```diff
@@ -28,44 +28,37 @@
 #include "entry.h"
 
 #include "factory/plugin-factory.h"
-#include "factory/draft/plugin-invalidation.h"
-#include "factory/draft/preset-discovery.h"
+#include "factory/preset-discovery.h"
 
 #include "plugin.h"
 #include "plugin-features.h"
```
Trinitou, line 34:
```suggestion
```
is it really necessary to include it here? I mean, it's already included via `#include "factory/plugin.h"`.  Also e.g. `id.h` is also not included here

include/clap/clap.h
```diff
@@ -28,44 +28,37 @@
...
 #include "host.h"
```
Trinitou, line 35:
```suggestion
```
is it really necessary to include it here? I mean, it's already included via `#include "factory/plugin.h"`. Or is it for people to better find this header?

Trinitou (changes requested):

include/clap/universal-plugin-id.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+// Pair of plugin ABI and plugin identifier.
+//
+// If you want to represent other formats please send us an update to the comment with the
+// name of the abi and the representation of the id.
+typedef struct clap_universal_plugin_id {
+   // The plugin ABI name, in lowercase.
+   // eg: "clap", "vst3", "vst2", "au", ...
```
Trinitou, line 9: should we mention "null-terminated" here?

include/clap/universal-plugin-id.h
```diff
@@ -0,0 +1,26 @@
...
+   //   eg: "com.u-he.diva"
+   //
+   // AU: format the string like "type:subt:manu"
+   //   eg: "aumu:SgXT:VmbA"
+   //
+   // VST2: print the id as a signed 32-bits integer
+   //   eg: "-4382976"
+   //
+   // VST3: print the id as a standard UUID
+   //   eg: "123e4567-e89b-12d3-a456-426614174000"
```
Trinitou, line 24: should we mention "null-terminated" here?

include/clap/universal-plugin-id.h
abique: Yes, that'd be better.

include/clap/universal-plugin-id.h
abique: Yes as well.

include/clap/universal-plugin-id.h
Trinitou: see #387

include/clap/universal-plugin-id.h
Trinitou: see #387 as well

include/clap/clap.h
```diff
@@ -28,44 +28,37 @@
...
+#include "universal-plugin-id.h"
```
Trinitou, line 37: These includes can be removed I think which removes some clutter.
Created a PR: #388 (checks passed)

Trinitou (changes requested):

include/clap/universal-plugin-id.h
abique: yes

abique: On the other hand any uuid helper can deal with both lower case and upper case right?
It just has to be a valid uuid.

ChangeLog.md
```diff
@@ -1,3 +1,54 @@
...
+
+* [extension-id](conventions/extension-id.md): introduce some rules about extension ID naming.
+
+## Plugin Template
+
+* [plugin-template.c](src/plugin-template.c): implement thread-safe plugin entry init counter
+
+## Organization
+
+* `clap.h` no longer includes headers from `ext/draft` or `factory/draft`. Draft extension and factory headers must now be explicitly included, either individually or via the `draft.h` header.
```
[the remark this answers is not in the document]
abique: Done.

include/clap/universal-plugin-id.h
Bremmers: Yeah, requiring lower case can even cause problems if some people support uppercase anyway, and some don't.
Then perhaps add "Both lower and upper case are allowed.", so people know they have to be prepared for both?
It can be an issue if you compare strings.

PR #372, opened 2023-12-21
Expand a bit on the difference between init and create

baconpaul:
This documentation was by create, but it wasn't by init, so make it clear in both locations.


abique:
Thanks :+1:

0ax1:
Thanks Paul!

PR #375, opened 2023-12-28
added project load as explicit option

defiantnerd:
the context "loading a project/song" is the most prominent example of using a context information for plugin state persistence. it is also implemented in other audio standards.

I also changed the order of the constants using 1 for the "if in doubt" option.

### DIFF
--- a/include/clap/ext/state-context.h
+++ b/include/clap/ext/state-context.h
@@ -30,7 +30,7 @@
 extern "C" {
 #endif
 
-static CLAP_CONSTEXPR const char CLAP_EXT_STATE_CONTEXT[] = "clap.state-context.draft/2";
+static CLAP_CONSTEXPR const char CLAP_EXT_STATE_CONTEXT[] = "clap.state-context/2";
 
 enum clap_plugin_state_context_type {
    // suitable for loading a state as a preset
--- a/include/clap/ext/state-context.h
+++ b/include/clap/ext/state-context.h
@@ -10,7 +10,8 @@
 /// on the context.
 ///
 /// Briefly, when loading a preset or duplicating a device, the plugin may want to partially load
-/// the state and initialize certain things differently.
+/// the state and initialize certain things differently, like handling limited resources or fixed
+/// connections to external hardware resources.
 ///
 /// Save and Load operations may have a different context.
 /// All three operations should be equivalent:
@@ -20,6 +21,8 @@
 ///        clap_plugin_state_context.save(CLAP_STATE_CONTEXT_FOR_PRESET),
 ///        CLAP_STATE_CONTEXT_FOR_PRESET)
 ///
+/// If in doubt, choose CLAP_STATE_CONTEXT_FOR_PRESET as option.
+/// 
 /// If the plugin implements CLAP_EXT_STATE_CONTEXT then it is mandatory to also implement
 /// CLAP_EXT_STATE.
 
@@ -27,14 +30,18 @@
 extern "C" {
 #endif
 
-static CLAP_CONSTEXPR const char CLAP_EXT_STATE_CONTEXT[] = "clap.state-context.draft/1";
+static CLAP_CONSTEXPR const char CLAP_EXT_STATE_CONTEXT[] = "clap.state-context.draft/2";
 
 enum clap_plugin_state_context_type {
-   // suitable for duplicating a plugin instance
-   CLAP_STATE_CONTEXT_FOR_DUPLICATE = 1,
-
    // suitable for loading a state as a preset
-   CLAP_STATE_CONTEXT_FOR_PRESET = 2,
+   CLAP_STATE_CONTEXT_FOR_PRESET = 1,
+
+  // suitable for duplicating a plugin instance
+   CLAP_STATE_CONTEXT_FOR_DUPLICATE = 2,
+
+   // suitable for loading a state during loading a project/song
+   CLAP_STATE_CONTEXT_FOR_PROJECT = 3,
+
 };
 
 typedef struct clap_plugin_state_context {


abique:
So, `CLAP_STATE_CONTEXT_FOR_PROJECT` is equivalent to `clap_plugin_state->load()`?

include/clap/ext/state-context.h
```diff
@@ -20,21 +21,27 @@
...
-
    // suitable for loading a state as a preset
-   CLAP_STATE_CONTEXT_FOR_PRESET = 2,
+   CLAP_STATE_CONTEXT_FOR_PRESET = 1,
+
+  // suitable for duplicating a plugin instance
+   CLAP_STATE_CONTEXT_FOR_DUPLICATE = 2,
+
+   // suitable for loading a state during loading a project/song
+   CLAP_STATE_CONTEXT_FOR_PROJECT = 3,
```
abique, line 43: Maybe we can add that PROJECT is the identity operation?

include/clap/ext/state-context.h
```diff
@@ -20,21 +21,27 @@
...
+/// 
 /// If the plugin implements CLAP_EXT_STATE_CONTEXT then it is mandatory to also implement
 /// CLAP_EXT_STATE.
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-static CLAP_CONSTEXPR const char CLAP_EXT_STATE_CONTEXT[] = "clap.state-context.draft/1";
+static CLAP_CONSTEXPR const char CLAP_EXT_STATE_CONTEXT[] = "clap.state-context.draft/2";
```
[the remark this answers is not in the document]
defiantnerd: ok, will do

include/clap/ext/state-context.h
defiantnerd: the documentation said `CLAP_STATE_CONTEXT_FOR_PRESET` is equivalent to `clap_plugin_state->load()`. And it should be kept that way, since `CLAP_STATE_CONTEXT_FOR_PROJECT` has a deeper effect to how the plugin behaves, it does break connections to connected hardware devices and overrides existing ones.  Therefore it should not be default behavior, this should be PREST.

include/clap/ext/state-context.h
```diff
@@ -20,21 +21,27 @@
 ///        clap_plugin_state_context.save(CLAP_STATE_CONTEXT_FOR_PRESET),
 ///        CLAP_STATE_CONTEXT_FOR_PRESET)
 ///
```
Bremmers, line 23: I suggest we remove the part starting with "All three operations should be equivalent". Instead, we state that hosts are required to use this extension for all save and load operations if the plugin supports the extension.
Mixing the two state extensions is confusing I think.

abique: That doesn't work for host released before that extension.

defiantnerd: It also makes perfect sense to keep this as "FOR PRESET" since the PROJECT context will allow a full association with bound hardware resources. I would keep both extensions, one is for simpler plugins/hosts and projects, the second one for the advanced ones.

Bremmers: Sorry for my bad wording. My point is that a host should use just one of the state extensions when dealing with a particular plugin instance. Ideally it uses the new context one. If it does use the context extension it should use it for all saving and loading, rather than use the old one for some things and the new one for other things.

robbert-vdh (commented):
My opinion on this is still the same as before. `CLAP_STATE_CONTEXT_FOR_PROJECT` is equivalent to the regular usage of the state extension, so adding it effectively either deprecates the state extension, or duplicates the functionality. I think it's a good idea to avoid duplicating functionality unless absolutely required. In this case not having `CLAP_STATE_CONTEXT_FOR_PROJECT` means that both extensions can supplement each other. Most plugins won't need the state context, so they won't have to add support for it either. But if the functionality is duplicated, then now both plugins and hosts need to handle the situation where only one of the extensions is implemented. And worse, the behavior can be different between the old state extension and the `CLAP_STATE_CONTEXT_FOR_PROJECT` state context.

defiantnerd:
The extension is indeed replacing the original state extension. But I don't see another way for this, since it does not fit the purpose. But we can not deprecate the original state extension for now.

`CLAP_STATE_CONTEXT_FOR_PROJECT` is not the same as the simple state extension, since `FOR_PROJECT' is potentially more destructive. In my case the plugin instances are bound to specific unique hardwares (identified by serial number) and audio is also being output to external outs, delay compensated.

When loading a song, the order of plugin restoration can not be determined. Therefore the absolute assignment to a device is crucial when loading a project. But it is not giving a pleasent user experience, if the connection would be part of the state it can be ignored if the state is used to restore "in duplicate" context or "a preset loaded" context. The distinction of the context is important here and the functionality can not be provided by the original state extension.

I hope this clears things up a bit for you.

robbert-vdh:
`CLAP_STATE_CONTEXT_FOR_PROJECT` is how the state extension, and the default state handling in any other plugin API for that matter, is generally used by hosts. Your use case also fits the `CLAP_STATE_CONTEXT_FOR_PROJECT` usage of the regular state extension.

defiantnerd:
Again, in theory, you're right - but to make it work well in a good user experience, it must like I've proposed - I already handle it this way in VST3 and hosts we recommend support that feature ("adding the `Project` context to the loadable stream).

abique:
So from what I read, it appears to me that it is unclear how `state->save()` maps to the state context. And maybe this doesn't need to be specified and can remain ambiguous.

If someone wants to clear this ambiguity, he then uses the state context extension.

defiantnerd:
My original proposal does not need a context for the `save()` function. We can happily remove it. The actual idea is to know what to load and what to ignore from a state depending on the context. I am happy to change that back again.

baconpaul:
So I’m super late and not that invested here but…. If this is just adding a context to the state extension, why not take a state extension instance as an argument. 

That is, yu have something like “set context for (clap state *, context)” 

then you require the implementation of state by api not just by documentation. 

The host then does

set context for s
S->load
Context complete 

and the load function manages the state internally with a guard. 

Gives you the awkward state maintenance but avoids duplicated load and save api points

this may be a terrible idea and ignore it if it is

abique:
@baconpaul the decorator api was suggested in the past but I didn't like it.
Because, you may not see it at all and it requires the plugin to cache the context and you never know what mistake can happen.
While on the other hand, passing the context to the load/save is a coding style I prefer.

baconpaul:
Yeah that makes sense
My thinking was if the save ctx to a pointer to just the regular state extension it would force the constraint that you must implement both and avoid redundancy but at the cost of a non atomic api

abique:
I think that the duplication of the `save()` method isn't the end of the world, I agree it isn't great but it isn't hard either to deal with.

What `save()` should do in regard to the context, is ambiguous Tim thought `PRESET` was best, I thought `PROJECT` was better and at the end of the day, the plugin decides what it wants to save and what works best by default for its particular needs. I think it is totally fine to not specify to which context `save()` maps to.

abique:
Do we have a consensus?

Bremmers:
I would suggest you accept this PR and update the comments.
AFAICS only Robbert isn't too happy with this. I used to agree with him, but it has become clear to me that using the context extension for two states and the normal state extension for one state is confusing and will lead to problems.

PR #382, opened 2024-01-15
Expand note expression tuning description just a bit

baconpaul:
1. fix a small grammar mistake
2. crystal clear on what semitone means


baconpaul:
This is already in the changelog; just a tweak based on a review that came in after you merged @abique

abique (changes requested):
I think after `;' it shouldn't be capitalized but then the sentence should terminate with `.' maybe because it already used punctuation?

abique:
I take care of it.

PR #384, opened 2024-01-17
Add CLAP_CONSTEXPR to voice-info and preset-load id declarations

baconpaul:
To be consistent with all the other extensions


abique:
Thanks

PR #388, opened 2024-01-18
Remove unnecessary includes from clap.h

Trinitou:
let's see if this passes the pull-request workflows... 😉 

I think this removes some clutter. here is my theory:
- `plugin.h` is included via `factory/plugin-factory.h` and plugin extensions (makes sense to me)
  - `host.h` and `plugin-features.h` are included via `plugin.h` (makes sense as well)
- `universal-plugin-id.h` is included via `factory/preset-discovery.h` (where it is actually used)

### DIFF
--- a/include/clap/clap.h
+++ b/include/clap/clap.h
@@ -30,11 +30,6 @@
 #include "factory/plugin-factory.h"
 #include "factory/preset-discovery.h"
 
-#include "plugin.h"
-#include "plugin-features.h"
-#include "host.h"
-#include "universal-plugin-id.h"
-
 #include "ext/ambisonic.h"
 #include "ext/audio-ports-activation.h"
 #include "ext/audio-ports-config.h"


abique:
I'm not sure about that one.

The goal of `clap.h` is to pull everything except the draft.
If `x.h` is included by `y.h` and we decide to not add `x.h` to `clap.h`, you still get `x.h` **indirectly** by including `clap.h`.
**But**, if one day `x.h` isn't included anymore from `y.h`, then we need to remember to add back `x.h` to `clap.h`.

If we **directly** include all headers in `clap.h`, we don't rely upon **indirect** inclusion.

baconpaul:
If clap.h is meant to be universal header I agree it should include things directly even if they are also included transitively

Trinitou:
Should we close this (incomplete) pull-request as we want a more universal solution for all includes? Or should we leave it here and bring it into a 1.2.1 discussion later?

baconpaul:
Let’s change it to draft so we don’t forget

abique:
So what do we decide here?
I was not in favor of this change, but I don't mind either.

What's the last call? We close it?

Trinitou:
> So what do we decide here? I was not in favor of this change, but I don't mind either.
> 
> What's the last call? We close it?

I'm fine with closing

Issue #390, opened 2024-02-06
.DS_Store missing in .gitignore


NatureIsFrequency:
Clap repositories on MacOS currently flag .DS_Store files as changes to the repository

Suggest following the convention of adding .DS_Store to .gitignore

As per the following repos:
https://github.com/free-audio/clap-wrapper/blob/main/.gitignore
https://github.com/free-audio/clap-helpers/blob/main/.gitignore
https://github.com/free-audio/clap-juce-extensions/blob/main/.gitignore

PR #391, opened 2024-02-06
Fix: for issue: #390 Add .DS_Store to .gitignore

NatureIsFrequency:
Add .DS_Store to .gitignore (MacOS specific)

Fixes issue: #390 
".DS_Store missing in .gitignore"


baconpaul:
Oh @abique we should include this one! Can you edit it to point at next and we can sweep it into 1.2.2?

abique:
Thanks.

PR #394, opened 2024-02-19
Documentation: WildCards for PCK not allowed in note on

baconpaul:
The NoteOn event needs port/channel/key specified. Add this fact to the documentation.

### DIFF
--- a/include/clap/events.h
+++ b/include/clap/events.h
@@ -146,8 +146,8 @@ enum {
 //
 // In the case of note on events:
 // - The port, channel and key must be specified with a value >= 0
-// - A note-on event with a '-1' for Port Channel or Key is invalid and
-//   can be rejected or ignored by a plugin.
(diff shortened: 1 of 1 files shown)


include/clap/events.h
```diff
@@ -144,6 +144,12 @@ enum {
 // Well constructed plugins will search for voices and notes using
 // the entire tuple.
 //
+// In the case of note on events:
+// - The port, channel and key must be specified with a value >= 0
+// - A note-on event with a '-1' for Port Channel or Key is invalid and
+//   can be rejected or ignored by a plugin.
```
abique, line 150: It can also be rejected by the host then. The host can be consuming the plugin's note outupt.

baconpaul: Will update!

abique:
Thanks!

PR #415, opened 2024-07-17
Expand the Thread Doc to clarify and expand realtime

baconpaul:
Expand the thread-doc to clarify and expand realtime
    
Summarizing the conversation in discussion #414, this PR adds a more complete description of when the realtime constraint must be met, by which party, and how it interacts with the thread safe tag.


abique:
Thanks!

PR #419, opened 2024-09-02
CLAP 1.2.2

abique:


baconpaul:
I'm fine with 757a927 d9771f ce6f212 and 8323ec1

since I wrote 3c18216 I shouldnt' review it.

include/clap/ext/thread-check.h
```diff
@@ -17,23 +17,40 @@ extern "C" {
...
+///
+///    The real-time constraint on the [audio-thread] interacts closely with the render extension.
+///    If a plugin doesn't implement render, then that plugin must have all [audio-thread] functions
+///    meet the real time standard. If the plugin does implement render, and returns true when
+///    render mode is set to real-time or if the plugin advertises a hard realtime requirement, it
+///    must implement realtime constraints. Host provided [audio-thread] functions can be called
+///    from a plugin on the audio thread and must also therefore either meet the real-time constraint,
+///    not run plugins which advertise a hard realtime constraint via render or don't implement render,
+///    or expect potentially inconsistent rendering and performance from plugins when they re-enter the host.
+///
```
deeedob, line 49: This last long sentence is a bit hard to read. 

-  `and must also therefore` the 'also' is not needed
- `,not run plugins ...` also reads strange

I would suggest:

```
Host-provided [audio-thread] functions can be called from a plugin on the audio thread and must, therefore, either meet the real-time constraint, avoid running plugins that advertise a hard real-time constraint via render or don't implement render, or expect potentially inconsistent rendering and performance from plugins when they re-enter the host.
``` 

but IMO this sentence is still too long.

abique: @baconpaul ⬆️

include/clap/ext/draft/undo.h
```diff
@@ -1,8 +1,11 @@
 #pragma once
 
 #include "../../plugin.h"
+#include "../../stream.h"
 
-static CLAP_CONSTEXPR const char CLAP_EXT_UNDO[] = "clap.undo/2";
+static CLAP_CONSTEXPR const char CLAP_EXT_UNDO[] = "clap.undo/4";
+static CLAP_CONSTEXPR const char CLAP_EXT_UNDO_CONTEXT[] = "clap.undo_context/4";
+static CLAP_CONSTEXPR const char CLAP_EXT_UNDO_DELTA[] = "clap.undo_delta/4";
```
Trinitou, line 8: do you have the "/4" here to keep them in sync with the CLAP_EXT_UNDO ?

include/clap/ext/draft/undo.h
```diff
@@ -40,31 +43,29 @@ extern "C" {
...
-   uint64_t flags;
+   bool are_deltas_persistent;
 
-   // This represents the delta format version that the plugin is using.
+   // This represents the delta format version that the plugin is currently using.
    uint32_t format_version;
 } clap_undo_delta_properties_t;
 
-typedef struct clap_plugin_undo {
+// Use CLAP_EXT_UNDO_DELTA
```
Trinitou, line 66: this should be above the declaration of `clap_undo_delta_properties`, shouldn't it? Otherwise it looks a bit like `clap_undo_delta_properties` wasn't a part of `CLAP_EXT_UNDO_DELTA`

include/clap/ext/draft/undo.h
```diff
@@ -76,25 +77,43 @@ typedef struct clap_plugin_undo {
...
+                        size_t               delta_size);
+} clap_plugin_undo_delta_t;
+
+// Use CLAP_EXT_UNDO_CONTEXT
+// This is an optional interface, that the plugin can implement in order to know about
+// the current undo context.
+typedef struct clap_plugin_undo_context {
+   // Indicate if it is currently possible to perform a redo or undo operation.
+   // if can_* is false then it invalidates the corresponding name.
    // [main-thread]
```
Trinitou, line 105:
```suggestion
   // [main-thread & plugin-subscribed-to-undo-context]
```

include/clap/ext/draft/undo.h
```diff
@@ -76,25 +77,43 @@ typedef struct clap_plugin_undo {
...
-                                    uint64_t             flags,
-                                    const char          *undo_name,
-                                    const char          *redo_name);
-} clap_plugin_undo_t;
+   void(CLAP_ABI *set_can_undo)(const clap_plugin_t *plugin, bool can_undo);
+   void(CLAP_ABI *set_can_redo)(const clap_plugin_t *plugin, bool can_redo);
 
+   // Sets the name of the next undo or redo step.
+   // name: null terminated string.
+   // [main-thread]
```
Trinitou, line 111:
```suggestion
   // [main-thread & plugin-subscribed-to-undo-context]
```

include/clap/ext/draft/undo.h
```diff
@@ -112,35 +131,43 @@ typedef struct clap_host_undo {
...
-                               const void        *undo_delta,
-                               size_t             undo_delta_size);
+                               const void        *delta,
+                               size_t             delta_size,
+                               bool               delta_can_undo);
 
-   // Asks the host to perform the next undo step.
-   // This operation may be asynchronous.
+   // Asks the host to perform the next undo or redo step.
+   // This operation may be asynchronous and isn't available while the host is within a change.
```
Trinitou, line 168: is it a misbehaviour if the plugin calls `undo`/`redo` if it isn't available?

There are 2 cases:
- the plugin did not subscribe to context updates -> it cannot know if its available -> should be always ok if the plugin calls `undo`/`redo`
  - question: will the host ignore the request or will it potentially delay the undo until a running action has finished? or is this up to the host?
- the plugin is subsribed to context updates -> it should know and it could be considered a misbehaviour if it calls undo/redo

So maybe change the conditions to `[main-thread & (!plugin-subscribed-to-undo-context | can-undo)]` / `[main-thread & (!plugin-subscribed-to-undo-context | can-redo)]`

include/clap/ext/thread-check.h
abique: On this question, I prefer to let you discuss, I'm not very good at writing precise docs in correct English ;-)

include/clap/ext/draft/undo.h
```diff
@@ -76,25 +77,43 @@ typedef struct clap_plugin_undo {
    bool(CLAP_ABI *can_use_delta_format_version)(const clap_plugin_t *plugin,
```
Trinitou, line 77: should it be allowed to call this if `clap_undo_delta_properties_t::are_deltas_persistent` is false?

include/clap/ext/draft/undo.h
abique: We did some internal iteration with the /3 and I bumped it so we don't get confused with ABI mismatch. Also even if /3 wasn't released, there was still a /3 to /4 iteration internally.

include/clap/ext/draft/undo.h
abique: Here we have a few interfaces, so we've explicitly indicated which extension id to use to get the struct.

include/clap/ext/draft/undo.h
abique: When the host is within a change, it simply set can_undo and can_redo to false.

abique: The plugin can use the host interface without implementing any of the plugin interfaces, so it is OK to call undo/redo without checking can_undo/redo. The host will simply ignore it in worst case.

include/clap/ext/draft/undo.h
abique: 🤔 maybe not

include/clap/ext/draft/undo.h
abique: Ah, and yes I kept the same number in all of them so they are in sync.

include/clap/ext/draft/undo.h
abique: Maybe I should call it request_undo() and request_redo() instead? :thinking:

abique: To conclude this discussion, I'd not make it a misbehavior, because it'll make the interface more complicated to check and impl correctly.
Think that the plugins may be running in another process, with the gui on another thread, and all the state changes needs to be communicated, it is all asynchronous and it is maybe better to be a bit relaxed here.

include/clap/ext/draft/undo.h
abique: Adding this requirement could be correct, but I don't see it necessary.
I prefer to not add this requirement to the spec, helpers, ...

include/clap/ext/draft/undo.h
```diff
@@ -112,35 +130,46 @@ typedef struct clap_host_undo {
...
-                               size_t             redo_delta_size,
-                               const void        *undo_delta,
-                               size_t             undo_delta_size);
+                               const void        *delta,
+                               size_t             delta_size,
+                               bool               delta_can_undo);
 
-   // Asks the host to perform the next undo step.
-   // This operation may be asynchronous.
+   // Asks the host to perform the next undo or redo step.
```
Trinitou, line 166: + "The host may ignore the request if it has no undo/redo step to perform." ?

abique: Thanks @Trinitou for all your useful feedback.
I added your suggestion and renamed to request_undo/redo so it is clearer that the host can discard the request.

Bremmers (approved):

include/clap/ext/thread-check.h
baconpaul: Here's a cleaner version. Original

> Host provided [audio-thread] functions can be called
> from a plugin on the audio thread and must also therefore either meet the real-time constraint,
> not run plugins which advertise a hard realtime constraint via render or don't implement render,
> or expect potentially inconsistent rendering and performance from plugins when they re-enter the host.

Updated

> Hosts also provide functions marked [audio-thread]. These can be safely called by a plugin in the audio 
> thread. Therefore hosts must either (1) implement those functions meeting the real-time constraints or
> (2) not process plugins which advertise a hard realtime constraint or don't implement the render extension. 
> Hosts which provide [audio-thread] functions outside these conditions may experience inconsistent or 
> inaccurate rendering.

thoughts?

deeedob: Sounds good to me!

abique: I copy the updated text then, thanks!

abique: Done.

abique:
Now that all the comments are addressed, I'll squash the history a bit so it is ready for merging.

deeedob (approved):

PR #420, opened 2024-09-11
Update undo draft

abique:


tim-janik:
Glancing over this draft, I have two questions:

1) When undo/redo processing is left to the host, how does the host learn that a new undo step should be added to the undo history, in case a user makes a change purley through the plugin UI? Consider hosts that continuously display a list of recorded undo history (e.g. like gimp's "Undo History" tab).

2) I am not sure the API allows the host to provide a fully linearized undo/redo history as described in [TheGURQ](https://github.com/zaboople/klonk/blob/master/TheGURQ.md). Do you have thoughts on that? https://github.com/zaboople/klonk/blob/master/TheGURQ.md

abique:
> Glancing over this draft, I have two questions:
> 
>     1. When undo/redo processing is left to the host, how does the host learn that a new undo step should be added to the undo history, in case a user makes a change purley through the plugin UI? Consider hosts that continuously display a list of recorded undo history (e.g. like gimp's "Undo History" tab).

The plugin calls `host->begin_change()` and `host->change_made()` or `host->cancel_change()`.

>     2. I am not sure the API allows the host to provide a fully linearized undo/redo history as described in [TheGURQ](https://github.com/zaboople/klonk/blob/master/TheGURQ.md). Do you have thoughts on that? https://github.com/zaboople/klonk/blob/master/TheGURQ.md

It is fully linearized.

PR #423, opened 2024-10-15
Adding draft extension for host-provided scratch memory

jatinchowdhury18:


abique:
Hi @jatinchowdhury18 

I have some questions with the interface.
1. Is this scratch memory thread-local?
2. Can you get the scratch memory pointer from the thread pool? If so is it shared amongst the all the tasks, or each one will use the current thread local one?

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include "../../plugin.h"
+
+// This extension lets the plugin request "scratch" memory
+// from the host. Scratch memory can be accessed during the
+// `process()` callback, but is not persistent between callbacks.
```
abique, line 7: It'd be great to have a paragraph that explain what are the benefits of using this extension.
I suppose that you were thinking of sharing the scratch buffer between multiple plugin instances?

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,43 @@
...
+
+   // Asks the host for the previously requested scratch memory.
+   // If the host returned "true" when scratch memory was requested,
+   // then this method must return a pointer to a memory block at least
+   // as large as the requested size. If the host returned "false"
+   // when scratch memory was requested then this method should return
+   // null.
+   //
+   // Note that this memory may be uninitialized, so the plugin may
+   // want to "zero" the memory before using it.
```
abique, line 35: I'd change the last paragraph to something like:

The memory isn't initialized and may have been used by other plugin instances before. So the plugin **must** correctly initialize the memory when using it.

abique (changes requested):

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,43 @@
...
+   // Asks the host for certain amount of scratch memory.
+   // If the host is unable to provide the memory, it should
+   // return "false".
+   //
+   // Note that any memory the host allocates to satisfy
+   // the requested scratch size can be de-allocated
+   // when the plugin is de-activated.
+   //
+   // [main-thread & being-activated]
+   bool(CLAP_ABI *request_size)(const clap_host_t *host, size_t scratch_size_bytes);
```
baconpaul, line 25: We have a bunch of memory pool stuff in short circuit and for this api I used the name 'pre-reserve' rather than 'request' to make it clear that this is something you do *before* your process loop as part of activation. Since the API seems to be 'i will ask for the block later' type thing.

include/clap/ext/draft/scratch-memory.h
defiantnerd: > I'd change the last paragraph to something like:
> 
> The memory isn't initialized and may have been used by other plugin instances before. So the plugin **must** correctly initialize the memory when using it.

I find the word *must* misleading. It implies that the plugin shall zero it before use - it just has to expect that it is not initialized.

defiantnerd:
> Hi @jatinchowdhury18
> 
> I have some questions with the interface.
> 
> 1. Is this scratch memory thread-local?
> 2. Can you get the scratch memory pointer from the thread pool? If so is it shared amongst the all the tasks, or each one will use the current thread local one?

The only constraint is that the memory is valid for a plugin **during** it's `process` call. Like the event queue or the audio buffers. No further lifetime guarantees are given.

include/clap/ext/draft/scratch-memory.h
abique: To me, initialize just means write before read, it doesn't imply to fill the memory with 0 first.

abique: Though it can be interesting for a glitch plugin to use uninitialized scratch memory ;-)

abique:
> > Hi @jatinchowdhury18
> > I have some questions with the interface.
> > 
> > 1. Is this scratch memory thread-local?
> > 2. Can you get the scratch memory pointer from the thread pool? If so is it shared amongst the all the tasks, or each one will use the current thread local one?
> 
> The only constraint is that the memory is valid for a plugin **during** it's `process` call. Like the event queue or the audio buffers. No further lifetime guarantees are given.

Then consider this:
- each voice needs 10K of scratch buffer
- you process 32 voices in parallel

If you have a global scratch buffer, you can't process your voice in parallel.
If you have a thread local scratch that you can query from the thread pool, then you can process your voices in parallel.

I believe this needs to be clarified in the spec.

defiantnerd:
> 
> Then consider this:
> 
> * each voice needs 10K of scratch buffer
> * you process 32 voices in parallel
> 
> If you have a single scratch buffer for process(), you can't process your voice in parallel, or you need 32 * 10K. If you have a thread local scratch that you can query from the thread pool, then you can process your voices in parallel with a scratch buffer of 10K.
> 
> I believe this needs to be clarified in the spec.

You can only request/register one buffer per instance with a predefined size during activation. The host signals in the return value of `request_size` if you can access that.

In the process call you just call `auto *scratchmem = ext_scratch_memory->access(host);` and if `scratchmem` is not a null you can just use it during the process call.

But there is only one buffer and the size does not change as long as the plugin is active.

baconpaul:
> > Then consider this:
> > 
> > * each voice needs 10K of scratch buffer
> > * you process 32 voices in parallel
> > 
> > If you have a single scratch buffer for process(), you can't process your voice in parallel, or you need 32 * 10K. If you have a thread local scratch that you can query from the thread pool, then you can process your voices in parallel with a scratch buffer of 10K.
> > I believe this needs to be clarified in the spec.
> 
> You can only request/register one buffer per instance with a predefined size during activation. The host signals in the return value of `request_size` if you can access that.
> 
> In the process call you just call `auto *scratchmem = ext_scratch_memory->access(host);` and if `scratchmem` is not a null you can just use it during the process call.
> 
> But there is only one buffer and the size does not change as long as the plugin is active.

I think Alex's point is if you use the thread pool extension to schedule jobs, those thread pool local jobs will be effectively parallel and running under process.

Can a thread-pool extension job access memory? If so is it distinct per thread or is it a single memory location?

My guess is: The thread pool and memory scratch extensions need some careful co-consideration. And the patterns where people use the scratch memory will also require scratch-per-thread-voice not scratch-per-process-block in those cases.

jatinchowdhury18:
Yeah, thanks for bringing this up, I hadn't considered the inter-operation of scratch-memory and the thread-pool extension.

I think the simplest solution would be to have the plugin request the total amount of scratch memory that it needs across all possible threads. Since `thread_pool.request_exec` is blocking, I think it should be safe for the thread pool jobs to use the scratch memory, provided the plugin makes sure that each thread is using an independent "chunk" of the memory. So a simple example would look like:

```cpp
struct My_Plugin
{
    size_t scratch_mem_per_voice = 10'000;
    size_t num_voices = 32;
    char* scratch_memory_data = nullptr;

    void activate(...) {
        scratch_memory_ext.pre_reserve(host, scratch_mem_per_voice * num_voices);
    }

    void process(...) {
        // Get all the scratch memory here
        scratch_memory_data = scratch_memory_ext.access(host);
        thread_pool_ext.request_exec(host, num_voices);
    }

    void thread_pool_callback(uint32_t task_index) {
        // Get a partition of the scratch memory for this voice to use
        char* this_voice_scratch_mem = scratch_memory_data + scratch_mem_per_voice * task_index;
        // do the actual DSP work here...
    }
};
```
However, I can see a few reasons why this might not be an ideal solution... For example, if the host thread pool only has 8 threads, then reserving enough scratch memory for 32 voices to be processed in parallel is a bit wasteful. Maybe the best solution is to have two scratch memory mechanisms... one scratch buffer that is intended to be accessed during the `process()` callback, and another to be accessed during the `thread_pool_exec()` callback?

```c
typedef struct clap_host_scratch_memory {
    bool(CLAP_ABI *pre_reserve_process_scratch)(const clap_host_t *host, size_t scratch_size_bytes);
    void*(CLAP_ABI *access_process_scratch)(const clap_host_t *host);

    bool(CLAP_ABI *pre_reserve_thread_pool_scratch)(const clap_host_t *host, size_t scratch_size_per_thread_bytes);
    void*(CLAP_ABI *access_thread_pool_scratch)(const clap_host_t *host, uint32_t task_index);
} clap_host_scratch_memory_t;
```

All that said, I don't have much experience with the thread-pool extension, so I'll defer to your more informed opinions :).

abique:
Here is my proposal:

The scratch is a thread local pointer, so if you retrieve it from the process call, you'll get a pointer that you can share with all the jobs. If you retrieve it from the thread pool, you get a pointer that is only for the current job.

I think this is the correct direction because it corresponds to how the host will implement this feature: each audio thread will have a single scratch buffer (thread local) whose size is greater or equal to the max requested size of all plugin instances.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -18,11 +34,11 @@ typedef struct clap_host_scratch_memory {
...
    //
    // Note that any memory the host allocates to satisfy
-   // the requested scratch size can be de-allocated
-   // when the plugin is de-activated.
+   // the requested scratch size can be de-allocated by the
+   // host when the plugin is de-activated.
    //
    // [main-thread & being-activated]
-   bool(CLAP_ABI *request_size)(const clap_host_t *host, size_t scratch_size_bytes);
+   bool(CLAP_ABI *pre_reserve)(const clap_host_t *host, size_t scratch_size_bytes);
```
abique, line 41: Maybe just call it `reserve`?

This function is not as easy as it seems to impl, as for returning true, it implies that the memory is successfully reserved. For bitwig, this means, allocating the memory for each threads in the thread pool, sending the new memory to those threads, in a way which guarantees that the memory will be setup before the plugin is processed.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,63 @@
...
+   // Asks the host for certain amount of scratch memory.
+   // If the host is unable to provide the memory, it should
+   // return "false".
+   //
+   // Note that any memory the host allocates to satisfy
+   // the requested scratch size can be de-allocated by the
+   // host when the plugin is de-activated.
+   //
+   // [main-thread & being-activated]
+   bool(CLAP_ABI *pre_reserve)(const clap_host_t *host, size_t scratch_size_bytes);
```
Trinitou, line 41: Do we ever use `size_t` in any other extension? -> `uint?_t` ?

abique: Good point, `uint32_t` would do the job I think.

baconpaul: Why not use uint64_t which is size_t in most of our systems? The host has the option to return no for values out of bounds

jatinchowdhury18: I don't have a preference for `uint32_t` vs `uint64_t`... but Trinitou is right that CLAP doesn't use `size_t` anywhere else, so I don't think we should use it here.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,63 @@
...
+#endif
+
+typedef struct clap_host_scratch_memory {
+   // Asks the host for certain amount of scratch memory.
+   // If the host is unable to provide the memory, it should
+   // return "false".
+   //
+   // Note that any memory the host allocates to satisfy
+   // the requested scratch size can be de-allocated by the
+   // host when the plugin is de-activated.
```
Trinitou, line 38: When exactly is the last time that the plugin is allowed to access the scratch memory? While it might be obvious, it's maybe good to write that down explicitly (?)

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: `reserve` is good!

Yeah, I think this extension will be tricky for hosts that process plugins across multiple threads, but for single-plugin-thread hosts I don't think it will be too complicated. I imagine this extension will be most appealing to hosts running on memory-constrained platforms, and I would guess most of those will have a single plugin thread. At least that's my use case at the moment :).

jatinchowdhury18:
> The scratch is a thread local pointer, so if you retrieve it from the process call, you'll get a pointer that you can share with all the jobs. If you retrieve it from the thread pool, you get a pointer that is only for the current job.

This makes sense! I've added some comments intended to clarify this point, but please let me know if there are ways I can improve my explanation :).

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: The plugin is only allowed to access the scratch memory during the `process()` callback... you're right that should be made more explicit.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,71 @@
...
+   // In the context of plugins and hosts that implement
+   // the "thread-pool" extension, scratch memory is assumed
+   // to be "thread-local". The plugin should request the maximum
+   // amount of scratch memory that it will need on a single
+   // thread. Accordingly, the host must ensure that each
+   // thread can independently provide the requested amount
+   // of scratch memory.
+   //
+   // [main-thread & being-activated]
+   bool(CLAP_ABI *reserve)(const clap_host_t *host, uint32_t scratch_size_bytes);
```
Trinitou, line 49: Maybe one more question: may the plugin ask multiple times (with decreasing scratch_size_bytes) as long as the host returns false?

jatinchowdhury18: I think that should be allowed... maybe if the host rejects the amount of scratch memory that the plugin initially requests, the plugin can opt for a more memory-efficient algorithm, or something like that.

include/clap/ext/draft/scratch-memory.h
abique: I wonder if it wouldn't be better if the host was pulling the scratch size :thinking:

include/clap/ext/draft/scratch-memory.h
abique: A scratch size bigger than 4 GB would be problematic I suppose.
Anyway regardless of the type, many host will likely have their own threshold.

include/clap/ext/draft/scratch-memory.h
abique: @Trinitou maybe it is too strict?

I'm fine with the plugin asking as many times as it wants.
The last call "wins", meaning that if the last call failed, no memory is reserved, if the last call succeed then the reserved memory is the one from the last call.

Though, let's add a comment saying that this function is expensive so the plugin should avoid unnecessary calls.

Trinitou: > Though, let's add a comment saying that this function is expensive so the plugin should avoid unnecessary calls.

I agree, you're right. That can work as well, totally. 👍 

I just thought that maybe it's more clear (and misbehaviour-checkable) to have more strict constrains. E.g. a plugin validator would have to set an arbitrary limit of `reserve`-calls to detect plugins asking too often.

Maybe both? Strict limitation + comment saying that this function is expensive so the plugin should avoid unnecessary calls

But as I said, I'm fine with both

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,79 @@
...
+
+   // Asks the host for the previously requested scratch memory.
+   // If the host returned "true" when scratch memory was requested,
+   // then this method must return a pointer to a memory block at least
+   // as large as the requested size. If the host returned "false"
+   // when scratch memory was requested then this method should return
+   // null.
```
Trinitou, line 56:
```suggestion
   // Asks the host for the previously reserved scratch memory.
   // If the host returned "true" when scratch memory was requested,
   // then this method must return a pointer to a memory block at least
   // as large as the reserved size. If the host returned "false"
   // when scratch memory was requested, then this method must not
   // be called.
```
Just somehow it could be made more clear that it's related to the `reserve` function.

Also I think calling it when the host already said "no" makes no sense and just should be forbidden (?)

Trinitou (changes requested):

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: I think I'm against the "strict limitation", but this does bring up another point: The documentation needs to clarify that if the plugin calls `reserve()` multiple times, then the amount of available scratch memory will be the amount specified in the final `reserve()` call (assuming it returned true). I'll have to think about the best way to word that...

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: Thanks for the edits :). `must not be called` is maybe a little strict, but I think I'm in favor of it.

include/clap/ext/draft/scratch-memory.h
baconpaul: “The amount of memory available will have size of the last successful call to reserve”

abique: If the last call fails then it is 0.

abique: Do we need a call to release the reserved memory?

defiantnerd: > Do we need a call to release the reserved memory?
no - it is valid until deactivation. On activation, the plugin needs to reserve again.

Trinitou (approved):
It looks pretty good to me now. Any further details can be discussed on next IMO.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,90 @@
+#pragma once
+
+#include "../../plugin.h"
+
+// This extension lets the plugin request "scratch" memory
+// from the host. Scratch memory can be accessed during the
+// `process()` callback, but is not persistent between callbacks.
```
abique, line 7: "but it's content isn't persistent between callbacks."

It could be confused with it's allocation that isn't persistent.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,90 @@
...
+// in sequence, and each plugin requires 10K of scratch memory.
+// If each plugin pre-allocates its own scratch memory, then 30K
+// of memory is being allocated in total. However, if each plugin
+// requests 10K of scratch memory from the host, then the host can
+// allocate a single 10K scratch buffer, and make it available to all
+// three plugins.
+//
+// On memory-constrained platforms, this optimization may allow for
+// more plugins to be used simultaneously. On platforms with lots
+// of memory, this optimization may improve CPU cache usage.
```
abique, line 23: Both optimization: space and CPU cache are valid regardless of the available memory on the system.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,90 @@
...
+   //
+   // [main-thread & being-activated]
+   bool(CLAP_ABI *reserve)(const clap_host_t *host, uint32_t scratch_size_bytes);
+
+   // Asks the host for the previously reserved scratch memory.
+   // If the host returned "true" when scratch memory was requested,
+   // then this method must return a pointer to a memory block at least
+   // as large as the reserved size. If the host returned "false"
+   // when scratch memory was requested, then this method must not
+   // be called.
```
abique, line 67: then this method must not be called and will return `NULL`.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,90 @@
...
+   //
+   // In the context of plugins and hosts that implement
+   // the "thread-pool" extension, scratch memory is assumed
+   // to be "thread-local". The plugin should request the maximum
+   // amount of scratch memory that it will need on a single
+   // thread. Accordingly, the host must ensure that each
+   // thread can independently provide the requested amount
+   // of scratch memory.
+   //
+   // [main-thread & being-activated]
```
p--b, line 59: suggestion: I think `[being-activated]` should be documented somewhere, because I believe it is necessary for correctness here (assuming it means that the plugin may only call this whilst it is in the call stack of a call to `clap_plugin.activate` by the host): otherwise there could be a race between calls to `reserve` and `access`?

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,90 @@
...
+   // since the host implementation may be relatively expensive.
+   // If the plugin calls `reserve()` multiple times, then the
+   // final call determines the actual amount of scratch memory
+   // that will be available to the plugin. If the final call
+   // returns false then no scratch memory will be provided,
+   // regardless of any previous calls to `reserve()`.
+   //
+   // Note that any memory the host allocates to satisfy
+   // the requested scratch size can be de-allocated by the
+   // host when the plugin is de-activated.
```
p--b, line 49: question: does this imply that if a plugin transitions `activated -> deactivated -> being-activated`, in the final `being-activated` state it must re-request scratch memory via a call to `reserve`? Or is the prior reservation still valid?

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,90 @@
...
+   //
+   // In the context of plugins and hosts that implement
+   // the "thread-pool" extension, scratch memory is assumed
+   // to be "thread-local". The plugin should request the maximum
+   // amount of scratch memory that it will need on a single
+   // thread. Accordingly, the host must ensure that each
+   // thread can independently provide the requested amount
+   // of scratch memory.
```
p--b, line 57: thought: If I understand correctly, I think this could be challenging for a host to implement efficiently, as I'm not sure the host necessarily knows during the call to `reserve` how many concurrent jobs a plugin is going to request during `process`, and therefore how much memory needs to be allocated if it wants to be able to run every job in parallel. Consequently, the host would be forced to allocate the worst-case amount of requested memory for every thread pool thread it creates, which might be much _more_ memory than is actually required.

As an extreme example, if the plugin knows it is _never_ going to create thread pool jobs, then all of the per-thread scratch buffers allocated by the host for that plugin are wasted.

I wonder if it would be worth augmenting `reserve` to have a `uint32_t concurrent_buffers` argument, which;
- If `0`, means that all tasks receive the same buffer from `access` and it is up to the plugin to deal with that as it wishes (perhaps even a feature)
- If `n > 0` means that task indices up to `n - 1` will successfully receive a buffer pointer upon calling `access`, and task indices `n` and upwards may receive NULL?

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: > assuming it means that the plugin may only call this whilst it is in the call stack of a call to clap_plugin.activate by the host

I believe this is correct. I'm not sure where would be the right place to document the meaning of `being-activated` but I agree that we should.

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: I'm not sure I fully understand the state transition being described. The prior reservation is invalidated when the plugin is deactivated. The next time the plugin is activated it will have the opportunity to call `reserve()` again.

include/clap/ext/draft/scratch-memory.h
jatinchowdhury18: Adding a `concurrent_buffers` argument seems reasonable to me... @abique do you have any thoughts on this?

It would be nice if we could have that argument default to zero so plugins that don't implement `thread-pool` don't need to worry about it, but I know that's not possible for a C API.

include/clap/ext/draft/scratch-memory.h
abique: `[being-activated]` correspond to duration of `plugin->activate()`, when it returns the plugin is activated or deactivated.

You could describe that in `plugin.h` on the activate method.

include/clap/ext/draft/scratch-memory.h
abique: You never know as well on which thread of your thread pool you're going to call `process()`.
All your thread pool threads needs to have a thread local scratch memory that is `>=all_plugin_instance_max_scratch_size`.

abique: I'm not sure about `concurrent_buffers`.

At least it should be no more than a hint, and shouldn't change the behavior. Because we already have a clear semantic, where the plugin can fetch the scratch buffer before calling into the thread pool.
So for me, `0` means no hint.

The only use case for it would be that you are on a crazy modern CPU, like 192 arm machine, and you don't want to allocate 192 times the max scratch size, so you have some kind of pool of scratch buffers that you can pick and release into in a lock-free way 🤔 

I'm not sure, I think that a scratch size of up to 64KB is acceptable, and I'd simply allocate one for each thread and have a simple a robust thread local access, instead of a dodgy lock-free pool of scratch buffers.

If you have a 192 cores CPU, you probably have 32GB of ram if not more, so 192*64KB should be acceptable as well.

abique: Still, the `concurrent_buffers` doesn't convince me, because the goal is to share the buffer with other plugins.

If you have N plugins that says max M tasks, the num of scratch buffers that you need is `min(M*N, thread_pool_size)`.

The scratch buffer is the most beneficial when N grows, because it shares the most memory.
While `concurrent_buffers` is the most beneficial when N is small.

I don't see a way to take advantage of `concurrent_buffers` that is worth the pain.

p--b: > You never know as well on which thread of your thread pool you're going to call process().
> All your thread pool threads needs to have a thread local scratch memory that is >=all_plugin_instance_max_scratch_size.

I'm not sure that's necessarily true: it depends on how the host implements the thread pool. There are valid use cases that means that the host *does* know which threads are going to be used by which plugins.

p--b: > At least it should be no more than a hint, and shouldn't change the behavior

That would defeat the point: the motivation for the suggested change is to avoid the host being forced to allocate `max(any_plugin_scratch)` for _every_ thread pool thread it creates. The change allows the host to only allocate as many buffers as are actually going to be needed.

> and I'd simply allocate one for each thread and have a simple a robust thread local access

As a host you could still choose to do that - the `concurrent_buffers` request doesn't prevent the host from providing _more_ buffers than the plugin requests; in the way I worded it, tasks from index `n` upwards _may_ receive `NULL` (but _may_ still receive a buffer, if the host just chooses to allocate the max scratch for every thread in the system).

include/clap/ext/draft/scratch-memory.h
p--b: I think that ought to be documented, because the documentation for `access()` says:

```
   // Asks the host for the previously reserved scratch memory.
   // If the host returned "true" when scratch memory was requested,
   // then this method must return a pointer to a memory block at least
   // as large as the reserved size.
```

... making no mention of the impact of an interceding "deactivated" state.

I initially read this as meaning that the last call to `reserve` determined
the amount of memory available from `access` (always, irrespective of deactivations),
and the statement regarding deallocations meant that the host was allowed to
free / realloc the memory during the time the plugin was deactivated.

include/clap/ext/draft/scratch-memory.h
p--b: I think the key thing comes back to what I mentioned above RE the host needing to allocate memory for the threads which is not actually needed. Imagine the following flow where there is one homogeneous thread pool and the host has no concept of the number of concurrent scratch buffers are needed by a plugin.

1. Decide that you are going to allow for 8 audio worker threads. Create a thread pool with 8 threads.
2. Activate plugin A which requests `X` bytes of scratch memory. Allocate `(1+8)X` bytes of scratch memory; 1 for the audio thread, and one for each worker thread, and distribute scratch pointers as appropriate.
3. Start processing. Plugin A uses the 8 worker threads.
4. Some time later, activate plugin B which requests `Y > X` bytes of scratch memory.
5. The host must now reallocate the scratch buffer allocation to be `(1+8)Y` bytes, do the pointer juggling with the worker threads, and deallocate the old buffers.
6. Continue processing. Plugin A continues to use the 8 worker threads as before. Plugin plugin B, as it happens, doesn't need to use any worker threads.

Result: The host has allocated `(1+8)Y` bytes, whereas it only needed to allocate `Y + 8X`. If `Y >> X`, this becomes quite wasteful quite quickly.

If plugin B had been able to signal `concurrent_buffers == 0`:
* the total memory usage would be lower, as the host needs less memory allocated in step 5, and;
* if `Y <= (1 + 8)X`, the host can avoid doing any of the work in step 5 at all if it treats the entire scratch buffer initially allocated as one large arena.

---

If a host isn't interested in supporting any of these optimisations, they can just ignore the non-zero value of `concurrent_buffers` and allocate away merrily, so I don't see a real downside from the host's perspective?

jatinchowdhury18: I agree that there doesn't seem to be a down-side to informing the host about the number of `concurrent_buffers` the plugin might need, aside from the API being slightly more complex.

Even if the host may not have a way to use that information, I think it makes sense for the plugin to be as specific as possible about what it's asking for.

abique: @p--b thanks, I now get the idea behind your scenario and I see how it could be happening.

@jatinchowdhury18 then go for it! 👍

abique: @jatinchowdhury18 maybe you should add a comment that if the plugin asks for N concurrent buffers, but actually uses N+1, then what happens, crash? nullptr?

@p--b would `bool is_concurrent` suffice?

Trinitou: > It would be nice if we could have that argument default to zero so plugins that don't implement `thread-pool` don't need to worry about it, but I know that's not possible for a C API.

It is possible...by making it a separate extension. ;)

p--b: Sorry I missed your comment! Making it a `bool` would help signal to the host in the case where the plugin isn't going to use the threadpool at all, but doesn't unlock as many optimisations for the host if the plugin is using threadpool tasks: if the plugin knows it's only ever going to fire off 4 threadpool tasks, then it's potentially wasteful for the host to allocate memory into all N threadpool threads where it only needs to allocate into 4. Really I guess the ideal would be some way for the plugin to indicate how many parallel threadpool tasks it wants to execute _a priori_ as part of the threadpool extension, but without that, at least allowing it to be indicated here allows the host not to have to overallocate. That said, if it was `bool` or nothing, then `bool` is definitely better :)

PR #424, opened 2024-10-24
Fix abort on delete plugin from track FX.

Javantea:
The plugin-template.c example plugin code will crash Reaper when you delete it from the track FX because of the assertion made on line 422. This can be fixed by simply changing the inequality.


baconpaul (approved):
Yes this is clearly correct thank you!

abique:
Yes, this is correct! Thank you 👍

PR #426, opened 2024-10-30
Document the special lifetime of clap_event_midi_sysex::buffer

baconpaul:
per conversation on discord, oct 29

### DIFF
--- a/include/clap/events.h
+++ b/include/clap/events.h
@@ -302,14 +302,16 @@ typedef struct clap_event_midi {
 // call in which the event is delivered or (from plugin->host) only the
 // duration of a try_push call.
 //
-// Since `clap_output_events.try_push` requires you to make a copy of
-// an event, host implementers receiving sysex messages from plugin need
+// Since `clap_output_events.try_push` requires hosts to make a copy of
+// an event, host implementers receiving sysex messages from plugins need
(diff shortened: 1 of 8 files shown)


include/clap/events.h
```diff
@@ -297,11 +297,25 @@ typedef struct clap_event_midi {
...
+// clap_event_midi_sysex contains a pointer to a sysex contents buffer.
+// The lifetime of this buffer is (from host->plugin) only the process
+// call in which the event is delivered or (from plugin->host) only the
+// duration of a try_push call.
+//
+// Since `clap_output_events.try_push` requires you to make a copy of
+// an event, host implementers receiving sysex messages from plugin need
+// to take care to both copy the event (so header, size, etc...) but
+// also allocate and memcpy the sysex pointer. Similarly plugins retaining
+// the sysex outside the lifetime of a single process call must copy the
```
Bremmers, line 309: This stuff runs in the audio thread, so you're not supposed to use the memory manager. "allocate" does suggest this I think?
If a receiver is only interested in small sysex messages this can easily be solved, but a general purpose receiver should probably pre-allocate some memory.

baconpaul: Yeah also copy to host owned memory is better. I’ll fix it up

PR #427, opened 2024-11-01
param-origin extension

Trinitou:
see in-code documentation. If something is not specified enough, please ask/propose changes of course 👍


baconpaul:
I commented that it seems this and the step resolution one could be combined into a single 'param-extended-info' with key/value which retained abi as we add keys.

PR #428, opened 2024-11-01
param-resolutions extension

Trinitou:
see in-code documentation. If something is not specified enough, please ask/propose changes of course 👍

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,5 @@ cmake-build*/
 
 # A place to store stuff and get it git ignored
 ignore/*
+
+.DS_Store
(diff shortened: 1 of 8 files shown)


baconpaul:
could these two be combined into a 'param-extended-info' extension which was also somehow extensible easily?

abique:
> could these two be combined into a 'param-extended-info' extension which was also somehow extensible easily?

I think it is cool that they have their own interface.
In the undo ext we did split the interface to lower the requirements on implementing one or the other.

abique:
Regarding the extension, in the semitone example, would the user wants to be able to "quantize" the steps?

So if he is at `0.23` he'd jump to `1.0` instead of `1.23`?
Would the host implicitly quantize or there would be a flag?
Toward which value should we quantize? Would it be the value origin?

PR #429, opened 2024-11-01
Add a param-origin extension

Trinitou:
Reopening #427

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -33,7 +33,7 @@ jobs:
         if: startsWith(matrix.os, 'macOS')
         run: |
           brew upgrade
-          brew install automake autoconf ninja cmake emscripten wasi-libc wasi-runtimes llvm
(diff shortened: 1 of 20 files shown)


messmerd:
This extension seems useful. What still needs to be done?

Trinitou:
> This extension seems useful. What still needs to be done?

Thank you for your interest. I think the main reason for not merging was that this is a per-parameter property and we couldn't agree whether a separate extension like this is ok or if it should be part of a params2 extension or some other mechanism for extended per-parameter metadata.

... I still would do it as proposed here.

abique:
Hey @Trinitou 

Please can you rebase this PR?
The added info looks fine to me, and the naming too.
Though it would be great to go through the discussion of naming we want for clap for this kind of extensions and perform the adjustments accordingly.
I would like to merge this one for 1.2.9.

Thank you very much :+1:

abique:
Ah github messed up the merge...

abique:
@Trinitou I managed to manually merge and squash your commits, so it is on the next branch!

PR #430, opened 2024-11-16
Webview extension (draft)

### LINKED ISSUE #416: Web GUI extension
I'm looking to build CLAP plugins as WASM bundles, where all the CLAP API calls are exported from the WASM module.  This looks (so far) like it would just work, which would be absolutely awesome.

One piece of the puzzle I haven't figured out the UI.  So far, my effects specify a page to open inside an `<iframe>`, and that web-page posts state updates which are translated into custom (non-CLAP) function calls.

## Option 1: dedicated extension

I would love to have an official CLAP extension to:
1. specify the path for a web-view UI
2. post/receive messages from that web-view (`clap_ostream_t` again?)
4. get notified about close/open

## Option 2: webview API type for `clap.gui`

This would mean adding a `clap_webview *webview` member to the `clap_window` union, and a `CLAP_WINDOW_API_...` constant.

The `clap_webview` would need to contain a (host-provided) method for posting messages, and registering a (plugin-provided) listener for incoming messages.  On the other hand, all the scaling/sizing stuff is already in `clap.gui`.

---

With either of these options, the plugin's `clap.gui` implementation (for native GUIs) could set up the webview and then translate messages to/from the other API.

geraintluff:
This draft lets CLAP plugins straightforwardly use a web-page for their UI.

Messages are then exchanged in both directions as opaque blocks of bytes.  The webpage receives and sends messages using standard web APIs.

The host is responsible for storing messages in suitable (lock- and allocation-free) queues.

Closes #416

### DIFF
--- a/include/clap/ext/draft/webview.h
+++ b/include/clap/ext/draft/webview.h
@@ -21,7 +21,7 @@ extern "C" {
 
 typedef struct clap_plugin_webview {
    // Returns the URL for the webview's initial navigation, as a null-terminated UTF-8 string.
-   // This must be called at least once before any messages are sent (or accepted by the host).
+   // This must be called while the webview is not open, and must be called before it is opened.
    // If this URL is relative, it is resolved relative to the plugin (bundle) resource directory.
    // The host may assume that no resources outside of that directory are used, and may use any
    // protocol to provide this content, following HTTP-like relative URL resolution. The page must
@@ -33,7 +33,7 @@ typedef struct clap_plugin_webview {
                              char                *out_buffer,
                              uint32_t            out_buffer_capacity);
 
-   // Receives a single message from the webview, which must be open and ready to receive replies.
+   // Receives a single message from the webview, which must be open.
    // Returns true on success.
    // [main-thread]
    bool(CLAP_ABI *receive)(const clap_plugin_t *plugin, const void *buffer, uint32_t size);
@@ -41,14 +41,15 @@ typedef struct clap_plugin_webview {
 } clap_plugin_webview_t;
 
 typedef struct clap_host_webview {
-   // Checks whether the webview is open and likely to receive messages.
-   // This may be called at any time and is a hint for non-essential work, such as being able to
-   // skip metering calculations.
+   // Returns whether the webview is "open" (running and ready to receive messages).
+   // If called from the main thread, it must return the open state specified by other methods.
+   // If called from any other thread, it must only be used as a hint for non-essential work (such
+   // as to skip metering calculations) as there are no synchronisation guarantees.
    // [thread-safe]
    bool(CLAP_ABI *is_open)(const clap_host_t *host);
 
    // Sends a single message to the webview.
-   // Returns true on success. It must fail (false) if the webview is not open.
+   // Returns true on success. Should fail if the webview is not open.
    // [main-thread]
    bool(CLAP_ABI *send)(const clap_host_t *host, const void *buffer, uint32_t size);
 


geraintluff:
Jules at ADC (Javascript, WebViews and C++)

Screenshot 2024-11-18 at 17 42 05

> At some point we're going to need to have a way for the DAW to ask the plugin: look, is the window you're giving me going to have a webview in it?  In which case just give me the HTML and we can put it in, we can actually embed it or put it in our own webview, or do something a bit less layer-y.

baconpaul:
I'm not a web tech person so this may be not useful.

shouldn't the file be called `webview.h` not `web.h` and similarly `clap_plugin_webview` etc....

all the functions are marked `[thread-safe]` which means they can be called from any function with any degree of concurrency. I'm surprised to see that vs `[main-thread]`. Why would the audio thread ever need to directly message the web view?

Shouldn't `get_start` have a more descriptive name like `provide_starting_uri`

the protocol for get start and is open are unclear. Is it possible to write a state diagram like the one in gui.h?

geraintluff:
This extension isn't specific to WCLAP (although it's useful there).  If we moved it to the other repo, might it get a different kind (subset?) of feedback?

geraintluff:
> shouldn't the file be called `webview.h` not `web.h` and similarly `clap_plugin_webview` etc....
✅ I agree with this.  `web.h` was initially suggested by @abique so I used that, but "webview" would be more accurate.

> Why would the audio thread ever need to directly message the web view?

My reasoning was: the UI should be updated about both state save/load (main thread) and parameter update events (audio thread).

The plugin could instead run its own message queue, post all the parameter updates there, request a main-thread callpack, and then forward them to the webview.  This way was easier for the plugin, the other way would be easier for the host. 🤷 
 
> Shouldn't `get_start` have a more descriptive name like `provide_starting_uri`

✅ Makes sense.

> the protocol for get start and is open are unclear. Is it possible to write a state diagram like the one in gui.h?

It should be possible to call either of them at any time.  I wasn't imagining either of them would trigger a state-change.

`is_open()` would return true if there was a webview UI running, so likely to receive messages sent to it.  So far, I've checked this on the state/parameter changes to see whether it should bother, which I guess means it'd only be called when the plugin is processing stuff?

baconpaul:
> My reasoning was: the UI should be updated about both state save/load (main thread) and parameter update events (audio thread).
> 
> The plugin could instead run its own message queue, post all the parameter updates there, request a main-thread callpack, and then forward them to the webview. This way was easier for the plugin, the other way would be easier for the host. 🤷

every plugin already does this to update their native GUI. We don't ever want to encourage gui writes from the audio thread vs 'get the message to a lock free queue' approach. (the reason is that every `[thread-safe]` function has to meet the full realtime constraints of no allocation, no network traffic, no disc i/o etc... and i don't think you can meet that here)

> It should be possible to call either of them at any time. I wasn't imagining either of them would trigger a state-change.

So can a host call `receive` on a plugin before it calls `get_start`? Or if `is_open` would return false? Is the plugin then supposed to preserve those messages until a start? (assuming these have better names in final)

geraintluff:
> So can a host call `receive` on a plugin before it calls `get_start`? Or if `is_open` would return false? Is the plugin then supposed to preserve those messages until a start? (assuming these have better names in final)

Really great points - I'll improve this and sketch out a diagram.

> > The plugin could instead run its own message queue [...]
> 
> every plugin already does this to update their native GUI. We don't ever want to encourage gui writes from the audio thread vs 'get the message to a lock free queue' approach. [...]

From the perspective of a (occasionally lazy 😛) plugin developer, that seems like wasted duplicate effort.  The host will _also_ have lock-free queues in its codebase already, and if the host got the message to its lock-free queue (instead of the plugin) then I thought perhaps the plugin wouldn't need to run such a queue at all, if it only supported the webview UI.  (And the `"web"` GUI type for sizing info)

I'm doubting that now though, especially since the plugin is much more likely to know how big the queue should be, what to do if the queue is full, etc.

Thanks for the challenging feedback!

baconpaul:
Yeah clap helpers has a lock free queue as does juce and other packages.  It’s already a paradigm plugins have. And frameworks wrap them if you want to think at a higher level. Also having a non allocating non blocking lock free data structure for a message of arbitrary size with multiple producers is …. Hard. And your api implies that a host has to implement that. Whereas the spsc fixed sized queues folks use to do interoperation between main and ui threads are well trodden and in almost every plugin codebase already

great stuff. Very interesting to see this evolve.

geraintluff:
@baconpaul I've updated things based on your comments.

The only method I left as `[thread-safe]` is `is_open()`, because (a) it's almost certainly just querying a boolean, and (b) I've clarified that when it should be used as a hint for non-essential stuff like metering, with no guarantees about synchronisation.

I personally think `is_open()` should either stay `[thread-safe]` or be removed completely.  If you really want to know whether the input is open, the `clap.gui` extension gives you explicit events (on the main thread) for that.  However, I found it very convenient to check from the audio for deciding whether to spend time collating/posting data for a main-thread metering message.

geraintluff:
Is there something else I should be doing to move this forward?

geraintluff:
Closing because of Git history chaos

PR #431, opened 2024-11-16
remove erroneous comment in ambisonic.h

interpunct:
it seems that the comment referred to a previous version of the ambisonic extension.


abique:
Thank you :+1:

PR #432, opened 2024-11-18
CLAP 1.2.3

abique:


include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,72 @@
...
+
+typedef struct clap_plugin_location {
+   // Called by the host when the location of the plugin instance changes.
+   //
+   // The last item in this array always refers to the device itself, and as
+   // such is expected to be of kind CLAP_PLUGIN_LOCATION_DEVICE.
+   // [main-thread]
+   void (*set_location)(clap_plugin_t                  *plugin,
+                        clap_plugin_location_element_t *path,
+                        int                             num_elements);
```
messmerd, line 67: Elsewhere in the CLAP API, `uint32_t` tends to be used for counts and indices.

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,72 @@
...
+
+static CLAP_CONSTEXPR const char CLAP_EXT_LOCATION[] = "clap.location/1";
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum {
+   // Represents a document/project/session.
+   CLAP_PLUGIN_LOCATION_PROJECT,
```
baconpaul, line 17: Elsewhere in CLAP we use explicit values for enums. So like `CLAP_PLUGIN_LCOATION_PROJECT=1` and so on.

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,72 @@
...
+   // Set to 0 if irrelevant.
+   uint32_t index;
+
+   // Internal ID of the element.
+   // This is not intended for display to the user,
+   // but rather to give the host a potential quick way for lookups.
+   const char *id;
+
+   // User friendly name of the element.
+   const char *name;
```
baconpaul, line 52: what's the lifetime of these const char *? Any reason we didn't use `char id[CLAP_TEXT_SIZE]` like we do most other places?

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,72 @@
...
+   clap_color_t color;
+} clap_plugin_location_element_t;
+
+typedef struct clap_plugin_location {
+   // Called by the host when the location of the plugin instance changes.
+   //
+   // The last item in this array always refers to the device itself, and as
+   // such is expected to be of kind CLAP_PLUGIN_LOCATION_DEVICE.
+   // [main-thread]
+   void (*set_location)(clap_plugin_t                  *plugin,
```
baconpaul, line 65: We need a `CLAP_ABI` decoration here don't we?

baconpaul (changes requested):
Added some comments on the location extension.

include/clap/ext/draft/location.h
abique: Yes.

include/clap/ext/draft/location.h
abique: Switching to char buffers, I think it makes it simpler you're right.

include/clap/ext/draft/scratch-memory.h
```diff
@@ -0,0 +1,91 @@
...
+   // In the context of plugins and hosts that implement
+   // the "thread-pool" extension, scratch memory is assumed
+   // to be "thread-local". The plugin should request the maximum
+   // amount of scratch memory that it will need on a single
+   // thread. Accordingly, the host must ensure that each
+   // thread can independently provide the requested amount
+   // of scratch memory.
+   //
+   // [main-thread & being-activated]
+   bool(CLAP_ABI *reserve)(const clap_host_t *host, uint32_t scratch_size_bytes);
```
p--b, line 61: question: Would you accept a PR that added a `concurrent_buffers` parameters as [discussed](https://github.com/free-audio/clap/pull/423#discussion_r1823552949) in #432?

defiantnerd (approved):

Bremmers (approved):

baconpaul:
LGTM also.

include/clap/ext/draft/scratch-memory.h
abique: @p--b ETA?

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,73 @@
...
+#endif
+
+enum {
+   // Represents a document/project/session.
+   CLAP_PLUGIN_LOCATION_PROJECT = 1,
+
+   // Represents a group of tracks.
+   // It can contain both track groups, tracks and devices (post processing).
+   // The first device within a track group has the index of
+   // the last track or track group within this group + 1.
```
messmerd, line 23: Specifically the first *device*, or should this be the first *child* (whether that is track group, track, or device?).

What is the purpose of setting the index to that value? Is it to tell the plugin the range of location elements that are children (direct or not) of the parent element? That should probably be made clear.

Is the first child required to immediately follow its parent element in the `clap_plugin_location_element_t *path` array in `set_location`?

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,73 @@
...
+   CLAP_PLUGIN_LOCATION_TRACK = 3,
+
+   // Represents a single device.
+   // It contains other nested device chains.
+   CLAP_PLUGIN_LOCATION_DEVICE = 4,
+
+   // Represents a nested device chain (serial).
+   // Its parent must be a device.
+   // It contains other devices.
+   CLAP_PLUGIN_LOCATION_NESTED_DEVICE_CHAIN = 5,
```
messmerd, line 36: Is there a reason why nested device chains & devices function differently from how track groups & tracks function? That is, why do device chains need to be a child of a device rather than function as a "device group" which can have child devices?

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,73 @@
...
+} clap_plugin_location_element_t;
+
+typedef struct clap_plugin_location {
+   // Called by the host when the location of the plugin instance changes.
+   //
+   // The last item in this array always refers to the device itself, and as
+   // such is expected to be of kind CLAP_PLUGIN_LOCATION_DEVICE.
+   // [main-thread]
+   void(CLAP_ABI *set_location)(clap_plugin_t                  *plugin,
+                                clap_plugin_location_element_t *path,
```
messmerd, line 67: If I understand correctly, this array is a flattened tree structure?

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,73 @@
...
+
+static CLAP_CONSTEXPR const char CLAP_EXT_LOCATION[] = "clap.location/1";
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum {
+   // Represents a document/project/session.
+   CLAP_PLUGIN_LOCATION_PROJECT = 1,
```
messmerd, line 18: Should this be required to be the first element in the array passed to `set_location`, if it is provided?

include/clap/ext/draft/location.h
```diff
@@ -0,0 +1,73 @@
...
+   CLAP_PLUGIN_LOCATION_NESTED_DEVICE_CHAIN = 5,
+};
+
+typedef struct clap_plugin_location_element {
+   // Kind of the element, must be one of the CLAP_PLUGIN_LOCATION_* values.
+   uint32_t kind;
+
+   // Index within the parent element.
+   // Set to 0 if irrelevant.
+   uint32_t index;
```
messmerd, line 45: So `index` does *not* refer to the array index of this element's parent element within the array passed to `set_location`?

I could be completely wrong, but it seems like a simpler way of flattening the tree would be to specify that the order of elements in `set_location`'s `path` array is the same order that elements of would be visited in a depth-first traversal of the tree. Then each `index` would be the array index of the direct parent. And since the root element does not have a parent, its `index` could be the index of the plugin device itself.

messmerd (commented):
Some questions and suggestions regarding the location extension

messmerd:
How much information should a host provide to a plugin through the location extension? Should it provide all the information it can, or only information about the plugin itself?

include/clap/ext/draft/location.h
abique: Yes it is an array, which describes the path toward the current plugin instance.

include/clap/ext/draft/location.h
abique: Actually, it could be simplified.
We could say that index starts at 0 for tracks and then 0 again for devices.

This is the model:

```
track_group
 `- track0 -|
 `- track1 -| -> (+) -> device0 -> device1 -> ... 
 `- ...    -|
```

include/clap/ext/draft/location.h
abique: A device chain is a serial list of devices.
A device can contain multiple device chains, which maybe use for parallel processing.
There are similarities between tracks and device chain. Though there's a different semantic, which the plugin can use for display. Think Fabfilter Pro Q3, with sidechain.

include/clap/ext/draft/location.h
abique: I don't think it should, because if your host is just a device chain, it doesn't have a track or a project.

abique:
> How much information should a host provide to a plugin through the location extension? Should it provide all the information it can, or only information about the plugin itself?

I'm not sure I understood well your question.

Think that you have a very large project, all you need to know your location is:
`/project/track_group[0]/track[3]/device[6]`

Here from the root track group, take the third track, then the 6th device and here we are.

Does that clarify the question?

include/clap/ext/draft/scratch-memory.h
abique: @p--b will you make that PR?

p--b: Hey, apologies for the delay. Yes, I can have something to you in the next 24 hours.

p--b: https://github.com/free-audio/clap/pull/436 :)

abique:
I've squashed some changes to the scratch-memory for `max_concurrency_hint`, and simplified the doc.

messmerd:
@abique 
Sorry it took so long for me to respond.

> Think that you have a very large project, all you need to know your location is:
/project/track_group[0]/track[3]/device[6]

Okay, I was assuming that since the location API allows the host to tell the plugin its index within a parent element (i.e. in your example, the device is on the 3rd track), it could also provide information about other parts of the project structure which are not directly related to the plugin itself.

For example, if a plugin is the 2nd effect in an effects chain, could the location API provide information about the effects before it and after it in the chain?

I was assuming it could, since I don't see what the purpose is in telling a plugin it is the 2nd effect in the chain if it cannot also know what devices are the 1st and the 3rd.

But if I'm wrong about that, as it seems that I am based on your comment, that would explain my initial confusion about how the location API is supposed to provide that information. All location information would be as simple as your example - it can only provide the minimal amount of information needed to identify a device's location within the project tree, similar to a filesystem path.

include/clap/events.h
```diff
@@ -9,7 +9,12 @@ extern "C" {
 #endif
 
 // event header
-// must be the first attribute of the event
+// All clap events start with an event header to determine the overall
+// size of the event and its type and space (a namespacing for types).
+// clap_event objects are contiguous regions of memory which can be copied
+// with a memcpy of `size` bytes starting at the top of the header. As
+// such, be very careful when desiginig clap events with internal pointers
```
messmerd, line 16:
```suggestion
// such, be very careful when designing clap events with internal pointers
```
typo

baconpaul:
I had suggested on discord and I think alex agreed that `gain_reduction` may be better as `gain_adjustment_display` since

1. It is an adjustment (+ve and -ve values)
2. It is a display thing the plugin reports, not a gain adjuster that the host applies to the signal

If 'display' bugs people then `gain_adjustment_reporting` is also a good choice, for hosts which don't display it.

abique:
Maybe I like `gain_adjustment_reporting` even more. :+1:

abique:
> Should the undo, scratch-memory, location, and gain-reduction extensions be listed in the README alongside the other extensions?

We can wait for those to be stable.

include/clap/ext/draft/gain-reduction.h
```diff
@@ -0,0 +1,35 @@
...
+  //
+  // The returned value is in dB. Zero means the plugin is applying no gain
+  // reduction, or is not processing. A negative value means the plugin is
+  // applying gain reduction, as with a compressor or limiter. A positive
+  // value means the plugin is adding gain, as with an expander. The value
+  // represents the dynamic gain reduction or expansion applied by the
+  // plugin, before any make-up gain or other adjustment. A single value is
+  // returned for all audio channels.
+  //
+  // [audio-thread]
```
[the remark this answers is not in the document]
baconpaul: One other thought: If this *is* an outbound event from the plugin, the plugin should also implement a 'start reporting' and 'stop reporting' audio thread method of some form to let it to know whether the host would listen to those events (and so they can be skipped for efficiency if needs be)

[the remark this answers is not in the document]
baconpaul: Happy with changes from schwa thanks for listening

include/clap/events.h
```diff
@@ -266,6 +271,12 @@ enum clap_transport_flags {
    CLAP_TRANSPORT_IS_WITHIN_PRE_ROLL = 1 << 7,
 };
 
+// clap_event_transport provides song position, tempo, and similar information
+// from the host to the plugin. There are two ways a host communicates these values.
+// In the `clap_process` structure sent to each processing block, the host must
```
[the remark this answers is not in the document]
abique: Transport is allowed to be null in the process struct, because some hosts may not have a transport, for example a mini host that runs a distortion effect in a guitar pedal.

include/clap/all.h
```diff
@@ -9,5 +9,7 @@
 #include "ext/draft/resource-directory.h"
 #include "ext/draft/transport-control.h"
 #include "ext/draft/triggers.h"
+#include "ext/draft/location.h"
```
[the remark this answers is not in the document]
abique: mmmh I'm not sure. It is not very important to me but if you think it is, then I can sort it.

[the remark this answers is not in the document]
abique: Though the stable ext ordering is important to me.

ChangeLog.md
```diff
@@ -1,3 +1,21 @@
+# Changes in 1.2.3
+
+## New draft extensions
+
+* [scratch-memory](include/clap/ext/draft/scratch-memory.h): host provided scratch memory within the process call
+* [location](include/clap/ext/draft/location.h): better info about the plugin location within the project
+* [gain-reduction](include/clap/ext/draft/gain-reduction.h): gain reduction feedback
```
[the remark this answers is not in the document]
abique: I've updated the changelog.

include/clap/all.h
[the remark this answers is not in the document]
abique: I've sorted them.

include/clap/events.h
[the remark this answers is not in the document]
baconpaul: Corrected in https://github.com/free-audio/clap/pull/443

schwaaa:
> Maybe I like `gain_adjustment_reporting` even more. 👍

I just saw this comment. Either way is fine with me, if you have a preference I'm happy to change it from the current gain_adjustment_display to gain_adjustment_reporting.

baconpaul:
> > Maybe I like `gain_adjustment_reporting` even more. 👍
> 
> I just saw this comment. Either way is fine with me, if you have a preference I'm happy to change it from the current gain_adjustment_display to gain_adjustment_reporting.

After our (painful, long) conversation yesterday I think the best name is `gain_adjustment_metering` since this is the first `metering` extension we have (a `metering` extension is an extension where you grab a value on the audio thread with a functionally const double get function at the end of a process block, since alex didn't read the entire conversation, appropriately)

PR #434, opened 2024-11-19
Add some of the feedback from the Auburn/DPlug port

baconpaul:
basically expand some comments around ranges, meaning, and so on.


abique:
Thanks 👍

PR #437, opened 2024-12-18
add zig bindings to language bindings section.

interpunct:
added my zig bindings to language bindings section. i host these on sr.ht, but if it would be preferred i also have a mirror here on github.


baconpaul:
Neat!

PR #440, opened 2025-01-08
adding ext/draft/gain-reduction.h

schwaaa:
Basic extension for plugin to report gain reduction to the host.

### DIFF
--- a/include/clap/ext/draft/gain-reduction.h
+++ b/include/clap/ext/draft/gain-reduction.h
@@ -11,17 +11,20 @@ extern "C" {
 #endif
 
 typedef struct clap_plugin_gain_reduction {
-  // Returns the current gain reduction in dB. the value is intended
+  // Returns the current gain reduction in dB. The value is intended
   // for informational display, for example in a host meter or tooltip.
   //
   // The function should be called on the audio thread to ensure consistency
   // with processing time, although the host will likely use the value on
   // the main thread.
   //
-  // The returned value is in DB. Zero means the plugin is applying no gain
+  // The returned value is in dB. Zero means the plugin is applying no gain
   // reduction, or is not processing. A negative value means the plugin is
-  // applying gain reduction. A positive value means the plugin is adding
-  // gain. A single value is returned for all audio channels.
+  // applying gain reduction, as with a compressor or limiter. A positive
+  // value means the plugin is adding gain, as with an expander. The value
+  // represents the dynamic gain reduction or expansion applied by the
+  // plugin, before any make-up gain or other adjustment. A single value is
+  // returned for all audio channels.
   //
   // [audio-thread]
   double(CLAP_ABI *get)(const clap_plugin_t *plugin);
--- a/include/clap/ext/draft/gain-reduction.h
+++ b/include/clap/ext/draft/gain-reduction.h
@@ -11,9 +11,20 @@ extern "C" {
 #endif
 
 typedef struct clap_plugin_gain_reduction {
-  // reports the current gain reduction in dB. the value is intended
-  // for informational display only, for example in a host meter or tooltip.
-   double(CLAP_ABI *current_gain_reduction_db)(const clap_plugin_t *plugin);
+  // Returns the current gain reduction in dB. the value is intended
+  // for informational display, for example in a host meter or tooltip.
+  //
+  // The function should be called on the audio thread to ensure consistency
+  // with processing time, although the host will likely use the value on
+  // the main thread.
+  //
+  // The returned value is in DB. Zero means the plugin is applying no gain
+  // reduction, or is not processing. A negative value means the plugin is
+  // applying gain reduction. A positive value means the plugin is adding
+  // gain. A single value is returned for all audio channels.
+  //
+  // [audio-thread]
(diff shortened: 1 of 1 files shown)


include/clap/ext/draft/gain-reduction.h
```diff
@@ -0,0 +1,21 @@
...
+static CLAP_CONSTEXPR const char CLAP_EXT_GAIN_REDUCTION[] = "clap.gain-reduction/0";
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct clap_plugin_gain_reduction {
+  // reports the current gain reduction in dB. the value is intended
+  // for informational display only, for example in a host meter or tooltip.
+   double(CLAP_ABI *current_gain_reduction_db)(const clap_plugin_t *plugin);
```
abique, line 16: Is it called on the audio thread after the process call?

schwaaa: I think either a main thread only or audio thread only model is reasonable, but my suggestion would be to explicitly document it as threadsafe. The processing time associated with the returned value is whatever the plugin thinks "current" means.

schwaaa: Probably best to document as well that the returned value is negative if the gain is reduced. Note that as written, this extension is essentially a copy of https://github.com/fenderdigital/presonus-plugin-extensions/blob/main/ipslgainreduction.h.

abique: I believe that the logical place to query that value is after the process call (audio-thread). Then the host can carry the value to its GUI itself.

You should add that this interface should only be implemented if the gain reduction is relevant for this plugin.

abique: What about renaming `current_gain_reduction_db()` to simply `get()`?

abique: Is stereo gain reduction relevant?

schwaaa: It's fine to document as audio thread only if you prefer, and renaming the accessor is fine. I'm not aware of any plugin that does stereo gain reduction which seems like it would be a highly specialized thing.

abique: You may want to mention that while the plugin isn't processing, the value can be assumed to be 0.

abique: Should we name the extension `clap_plugin_gain_reduction_feedback`? :thinking:

schwaaa: I think the extension name is fine, but if you want to change it, I would suggest clap_plugin_gain_reduction_reporting rather than _feedback.

Trinitou: I also think the extension name is fine.

Trinitou:
Should there be a fixed min/max dB value defined in a comment so that it is clear for the host what the min range needs to be in the case of drawing a gain reduction meter?

Trinitou:
If a plugin can change its type dynamically (e.g. if it is a compressor or an eq), should there be a mechanism to report that to the host?

schwaaa:
I don’t think that’s necessary. The plugin is either applying gain reduction or it’s not. On Jan 8, 2025, at 12:14 PM, Trinitou ***@***.***> wrote:﻿
If a plugin can change its type dynamically (e.g. if it is a compressor or an eq), should there be a mechanism to report that to the host?

Bremmers:
Not sure positive values should be allowed.
In our UI 0dB is on the right if it's a gain reduction meter. If it can boost too, 0dB would be at center position. But with the current design we don't know what kind of meter it is.
If positive values are allowed hosts are more or less forced to do something with it, or it won't work right with some plugins.
It would be easier to just stick to reduction.

abique:
Let's say it is a gain feedback instead.
So +3, means +3dB and -3 means -3dB.
In Bitwig, if the gain is negative, we could paint something in blue from top to bottom; if the gain is positive something in red from bottom to top.

That way the feedback would have a clear center value (0 for 0dB); it works both for reduction and expansion.
The extension would need to be renamed to `gain_feedback` or something like that to remove the possible confusion regarding the interpretation of the value direction.
What do you think?

schwaaa:
I think it's useful to support reporting positive gain adjustments. I think the name gain_reduction is fine even if positive adjustments are supported, as long as the documentation is clear, which I think it is currently. If the name is changed, I would suggest gain_adjustment or gain_adjustment_reporting instead of feedback, which imo has too many other meanings in an audio context.

Bremmers:
With the presonus (reduction only) idea it's clear it's about the dynamic reduction without any make-up gain. That's how I understand it at least. Because if the user turns make-up gain all the way up the level would be > 0dB all the time, and the meter wouldn't work.

If gain is allowed too, some plugins may include make-up gain. Now you're no longer sure what the meter means exactly. And the CLAP version of a plugin can be different than the VST3 version.

Perhaps we can add a comment like "the value is supposed to represent dynamic compression/expansion, not static gain like make-up gain."

schwaaa:
Good point about make-up gain. I've added to the documentation/comment in the header to clarify this.

abique:
Thanks :+1:

PR #443, opened 2025-01-18
The host *may* not *must* include a transport in process

baconpaul:
Clarify this clarification


baconpaul:
@abique I think we should pull this into next / 1.2.3

abique:
Thanks!
I've missed it, glad you pinged :)

PR #449, opened 2025-02-01
Add missing surround sound identifiers

### LINKED ISSUE #407: Clarification about surround extension
I have a question about the surround extension. 

Right now the surround extension provides an enum which includes top speakers, but it is unclear to me how they map exactly. The names `CLAP_SURROUND_TFC` and `CLAP_SURROUND_TBC` ("Front Center Height" and "Rear Center Height") suggest a speaker in front of and behind the listener but looking at the [dolby 7.1.6 specification](https://www.dolby.com/about/support/guide/speaker-setup-guides/7.1.6-overhead-speaker-setup-guide/) though they include (what would be) "Sides Left Height" and "Sides Right Height" speakers, which aren't included in the surround extension. 

Are these two sets of speakers the same, or are  they two different sets of speakers (one of which would then need to be added)?

messmerd:
This PR adds two missing surround sound identifiers: "Side Left Height" (or top middle left) and "Side Right Height" (or top middle right) as seen in the [Dolby 7.1.6 specification](https://www.dolby.com/about/support/guide/speaker-setup-guides/7.1.6-overhead-speaker-setup-guide/) which was mentioned in #407.

"Top Center" (`CLAP_SURROUND_TC`) already exists so I don't think a 3rd identifier is needed.
This addition should be backwards compatible.

As a disclaimer, I do not know anything about surround sound, so someone familiar with it should look over this PR.

Closes #407

### DIFF
--- a/include/clap/ext/surround.h
+++ b/include/clap/ext/surround.h
@@ -55,6 +55,8 @@ enum {
    CLAP_SURROUND_TBL = 15, // Rear Left Height
    CLAP_SURROUND_TBC = 16, // Rear Center Height
    CLAP_SURROUND_TBR = 17, // Rear Right Height
+   CLAP_SURROUND_TSL = 18, // Side Left Height
+   CLAP_SURROUND_TSR = 19, // Side Right Height
 };
 
 typedef struct clap_plugin_surround {
--- a/include/clap/ext/surround.h
+++ b/include/clap/ext/surround.h
@@ -41,11 +41,11 @@ enum {
    CLAP_SURROUND_FR = 1,   // Front Right
    CLAP_SURROUND_FC = 2,   // Front Center
    CLAP_SURROUND_LFE = 3,  // Low Frequency
-   CLAP_SURROUND_BL = 4,   // Back Left
-   CLAP_SURROUND_BR = 5,   // Back Right
+   CLAP_SURROUND_BL = 4,   // Back (Rear) Left
+   CLAP_SURROUND_BR = 5,   // Back (Rear) Right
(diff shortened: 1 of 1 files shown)


abique:
Thanks!

abique:
@roelofvkr maybe you want to approve this change?

roelofvkr:
Thanks for checking! This also looks good to me and the current spec be enough to support up to 9.1.6. 

I did see dolby has one more [spec for 11.1.8](https://www.dolby.com/siteassets/about/support/guide/setup-guides/11.1.8-mounted-overhead-speakers-setup-guide/sell-sheet-11.1.8-mounted2.pdf), which would require more additions (it seems like they have another set of rear speakers and another set of top rear/side speakers), but we don't support that and I couldn't really find any other software or plugin format that does. 

For me this is good, but I did want to mention the 11.1.8 spec for completeness' sake.

messmerd:
Do you think any of the identifier comments should be changed? (whether for this PR's additions or older identifiers)

I ask because there are abbreviations like "TSL" which stands for "Top Side Left", but the comment next to it says "Side Left Height". I don't know if there's a more official name they should be given in the comments.

abique:
> Do you think any of the identifier comments should be changed? (whether for this PR's additions or older identifiers)
> 
> I ask because there are abbreviations like "TSL" which stands for "Top Side Left", but the comment next to it says "Side Left Height". "Top"/"Height" and "Back"/"Rear" are used interchangeably.

I think they should be updated so they include both common descriptions.

abique:
Thank you very much, great work :-)

PR #450, opened 2025-02-01
Git Ignore Meson Generated Subproject Files

mfisher31:
This change adds a line to `.gitignore` to account for files generated when using Clap as a subproject within a meson project.


mfisher31:
canceling this one.  I screwed up my `meson` branch

PR #451, opened 2025-02-01
Ignore Meson Generated Files in Git

mfisher31:
When using a cmake project as a [meson](https://mesonbuild.com) subproject, meson generates `meson-subproject*.*` files.  This change ignores them.


abique:
Thanks :+1:

PR #454, opened 2025-02-05
Plugin State Converter rework

abique:


include/clap/factory/draft/plugin-state-converter.h
```diff
@@ -22,75 +22,121 @@ typedef struct clap_plugin_state_converter_descriptor {
...
-                               clap_id                             src_param_id,
-                               double                              src_plain_value,
-                               clap_id                            *dst_param_id,
-                               double                             *dst_plain_value);
+   // Note: one source parameter can map to many destination parameters, see the note above.
+   int32_t(CLAP_ABI *get_mapping)(struct clap_plugin_state_converter *converter,
+                                  clap_id                             src_param_id,
+                                  clap_id                            *dst_param_ids,
+                                  uint32_t                            dst_param_ids_size);
+
```
Bremmers, line 94: get_mapping is reduntant I think? convert_value can do this already.

include/clap/factory/draft/plugin-state-converter.h
```diff
@@ -22,75 +22,121 @@ typedef struct clap_plugin_state_converter_descriptor {
...
+   //
+   // Note:
+   // convert_value assumes the state of the plugin to the one from the most recent call to
+   // convert_state.
+   int32_t(CLAP_ABI *convert_value)(struct clap_plugin_state_converter        *converter,
+                                    uint32_t                                   flags,
+                                    clap_plugin_state_converter_param_value_t *src,
+                                    clap_plugin_state_converter_param_value_t *dsts,
+                                    uint32_t                                   dsts_size);
 } clap_plugin_state_converter_t;
```
Bremmers, line 113: I'd prefer to turn convert_state, get_mapping and convert_value into a single function:

   int32_t(CLAP_ABI *convert)(struct clap_plugin_state_converter        *converter,
                              uint32_t                                   flags,
                              const clap_istream_t               *src,
                              const clap_ostream_t               *dst,
                              clap_plugin_state_converter_param_value_t *srcs,
                              uint32_t                                   srcs_count,
                              clap_plugin_state_converter_param_value_t *dsts,
                              uint32_t                                   dsts_size);


This avoids the rather ugly "convert_value assumes the state of the plugin to the one from the most recent call to". It also gives plugin developers the whole shebang in one go, so we don't have to change it again when a developer needs to see more than one automation point to be able to do the right thing.

PR #455, opened 2025-02-06
mini-curve-display: new extension

abique:


baconpaul:
Oh what a great festure

my one bit of feedback is this https://xkcd.com/833/

although not every daw will use it I think having an api point so you can provide a title, x axis label and scale, and y axis label and scale may be useful for the future

PR #458, opened 2025-02-26
Add mini-curve-display-hints extension.

roelofvkr:
This extension allows the plugin to optionally query the host for hints as to how a curve of a certain kind will be displayed.

The idea is that all mini curves within the host can display within the same range to provide a more unified experience.


baconpaul:
I like this idea, but since mini curve is still in draft, any reason we can't consolidate this into mini curve for 1.3?

abique:
Yes I think this could go into the mini curve ext, I take care of it.
Thanks 👍

PR #460, opened 2025-04-02
Fix Spelling in entry.h

hoelzeli:
Found a spelling mistake, I guess that it should be "pattern" in the explanation.


abique:
Thank you :+1:

PR #461, opened 2025-04-10
Webview extension

geraintluff:
This draft lets CLAP plugins straightforwardly use a web-page for their UI.

Messages are exchanged in both directions as opaque blocks of bytes on the main thread.  The webpage receives and sends messages using standard web APIs.

### DIFF
--- a/include/clap/ext/draft/webview.h
+++ b/include/clap/ext/draft/webview.h
@@ -2,6 +2,10 @@
 
 static CLAP_CONSTEXPR const char CLAP_EXT_WEBVIEW[] = "clap.webview/1";
 
+// clap.gui API constant. The pointer in clap_window must be NULL, but sizing methods are useful.
+// uses logical size, don't call clap_plugin_gui->set_scale()
+static const CLAP_CONSTEXPR char CLAP_WINDOW_API_WEBVIEW[] = "webview";
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -17,6 +21,7 @@ extern "C" {
 
 typedef struct clap_plugin_webview {
    // Returns the URL for the webview's initial navigation, as a null-terminated UTF-8 string.
+   // This must be called at least once before any messages are sent (or accepted by the host).
    // If this URL is relative, it is resolved relative to the plugin (bundle) resource directory.
    // The host may assume that no resources outside of that directory are used, and may use any
    // protocol to provide this content, following HTTP-like relative URL resolution. The page must
@@ -24,11 +29,11 @@ typedef struct clap_plugin_webview {
    // absolute, including a `file://` URL.
    // Returns true on success.
    // [main-thread]
-   bool(CLAP_ABI *get_start)(const clap_plugin_t *plugin,
+   bool(CLAP_ABI *provide_starting_uri)(const clap_plugin_t *plugin,
                              char                *out_buffer,
                              uint32_t            out_buffer_capacity);
 
-   // Receives a single message from the webview.
+   // Receives a single message from the webview, which must be open and ready to receive replies.
    // Returns true on success.
    // [main-thread]
    bool(CLAP_ABI *receive)(const clap_plugin_t *plugin, const void *buffer, uint32_t size);
@@ -36,14 +41,15 @@ typedef struct clap_plugin_webview {
 } clap_plugin_webview_t;
 
 typedef struct clap_host_webview {
-   // Checks whether the webview is open (and ready to receive messages)
+   // Checks whether the webview is open and likely to receive messages.
+   // This may be called at any time and is a hint for non-essential work, such as being able to
+   // skip metering calculations.
    // [thread-safe]
    bool(CLAP_ABI *is_open)(const clap_host_t *host);
 
    // Sends a single message to the webview.
-   // It must not allocate or block if called from the audio thread.
-   // Returns true on success.
-   // [thread-safe]
+   // Returns true on success. It must fail (false) if the webview is not open.
+   // [main-thread]
    bool(CLAP_ABI *send)(const clap_host_t *host, const void *buffer, uint32_t size);
 
 } clap_host_webview_t;
--- a/include/clap/ext/draft/webview.h
+++ b/include/clap/ext/draft/webview.h
@@ -1,12 +1,12 @@
 #pragma once
 
-static CLAP_CONSTEXPR const char CLAP_EXT_WEB[] = "clap.web/1";
+static CLAP_CONSTEXPR const char CLAP_EXT_WEBVIEW[] = "clap.webview/1";
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-/// @page Web
+/// @page Webview
 ///
 /// This extension enables the plugin to provide the start-page for a webview UI, and exchange
 /// messages back and forth.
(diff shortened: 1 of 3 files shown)


include/clap/ext/draft/webview.h
```diff
@@ -0,0 +1,59 @@
...
+   // Returns the URL for the webview's initial navigation, as a null-terminated UTF-8 string.
+   // This must be called at least once before any messages are sent (or accepted by the host).
+   // If this URL is relative, it is resolved relative to the plugin (bundle) resource directory.
+   // The host may assume that no resources outside of that directory are used, and may use any
+   // protocol to provide this content, following HTTP-like relative URL resolution. The page must
+   // not assume that the root path of the domain is the root of the bundle. The URL may also be
+   // absolute, including a `file://` URL.
+   // Returns true on success.
+   // [main-thread]
+   bool(CLAP_ABI *provide_starting_uri)(const clap_plugin_t *plugin,
```
abique, line 32: maybe?
```C
// returns the uri length
int32_t (CLAP_ABI *get_uri)(const clap_plugin_t *plugin, char *uri, uint32_t uri_capacity);
```

include/clap/ext/draft/webview.h
```diff
@@ -0,0 +1,59 @@
...
+   bool(CLAP_ABI *receive)(const clap_plugin_t *plugin, const void *buffer, uint32_t size);
+
+} clap_plugin_webview_t;
+
+typedef struct clap_host_webview {
+   // Checks whether the webview is open and likely to receive messages.
+   // This may be called at any time and is a hint for non-essential work, such as being able to
+   // skip metering calculations.
+   // [thread-safe]
+   bool(CLAP_ABI *is_open)(const clap_host_t *host);
```
abique, line 48: This isn't needed maybe.
If I understand correctly this works together with the gui ext? Then the gui ext already tracks the window state.

include/clap/ext/draft/webview.h
```diff
@@ -0,0 +1,59 @@
...
+/// messages back and forth.
+///
+/// Messages are received in the webview using a standard MessageEvent, with the data in an
+/// ArrayBuffer. They are posted back to the plugin using window.parent.postMessage(), with the
+/// data in an ArrayBuffer or TypedArray.
+
+typedef struct clap_plugin_webview {
+   // Returns the URL for the webview's initial navigation, as a null-terminated UTF-8 string.
+   // This must be called at least once before any messages are sent (or accepted by the host).
+   // If this URL is relative, it is resolved relative to the plugin (bundle) resource directory.
```
abique, line 25: The concept of bundle isn't yet defined within clap, at least for web clap we haven't specified it yet.
This isn't a blocker for the draft, but we'll have to define it before stabilizing this ext.

geraintluff:
@abique I think the requirements-check is best done from inside the page.  Websites manage this OK (outside of ancient enterprise workplace tools which don't work in Firefox or whatever 😄), and modern web tech is pretty stable.

include/clap/ext/draft/webview.h
abique: Maybe you can allow a call with a null uri to just query the uri size.

geraintluff: For context: I was copying the params `value_to_text` method (down to the variable names) which similarly relies on null-terminating the string.  It seemed reasonable, and I also liked being able to return `false` for an error (e.g. the capacity is too small) instead of an empty string.

The host would also need to check that the returned length was less than the capacity - not a huge burden, but forgetting it would mean an otherwise-sandboxed WCLAP could get a URL like `path/to/ui.html#-----filler------[SOME PARTS OF HOST MEMORY]`.

(I know there's some skepticism about how well WCLAPs could be sandboxed _anyway_, and obviously any decent WCLAP bridge would have that length-check, but it's still something extra to stumble over.)

abique: There's a difference with parameters.
For example let's say you want to display the value on a small screen with 8 chars, then the plugin can decide upon a different formatting for small displays.

Here it isn't an option to be truncated.

abique: @geraintluff will you update this function's signature?

geraintluff:
@abique For the starting URI length, can we somehow fix the capacity at `CLAP_PATH_SIZE`?  That seems long enough, gives more consistency from hosts, and avoids an extra method in the API.

abique:
> @abique For the starting URI length, can we somehow fix the capacity at `CLAP_PATH_SIZE`? That seems long enough, gives more consistency from hosts, and avoids an extra method in the API.

For the draft we can, for the final version I'd need to think more about it.
Do the web browser have a max url size?
We could also define a `CLAP_URI_SIZE` if we want more space.

geraintluff:
I think we'd be better off setting a fixed size, because what is a plugin meant to do if the host gives it a smaller capacity?  The maximum length varies by browser/webview, but none of them are below 1024.

For context, this URL is 509 bytes - less than half the 1024 of `CLAP_PATH_SIZE`:
```
http://localhost:8000/tmp/web-audio/?url=/tmp/dynamap/web-audio/DynaMap.js&ui=/tmp/dynamap/web/#dynamap?preset=vVLBSsRADP2XnKNMlYr2qCx48aQ32UNsx93QdqY0abGU-XfZaWV3sUUQ9JSQeXnvJZMRetsKewfZDQI5qgZhgWw81q8QGtuyL55inXLl3kJmEHqqOgtZagICqVJe3nsvuowyl2lAqLwvaW9plWyGsSuX35MQEIrBUc35uc8E4W1oSNaIA0LetYfiSZNBaDw7FcheR_iA7OL6DmGYow5N7NweRLUlJ2wjdjxb20-yUnFuJ4XTtta-H7aQItRT2BFPc6h3E4nzLFNmC9ZnJf1mv-50QoivfEy-7JgQcFExmSWTf9RMzTym-QvNLYJ0opF3-WoQxDbUkka-RUy8vJnmJTpb-9Bo95F3-4du5dpvjTFHts2vhgqf&audio=audio%2Fexample.mp3
```

Local paths are relative to the plugin bundle, so can be short, and an absolute URL of 1024 bytes is long enough to hold quite a lot of config/auth-tokens/SHA-sums/etc.

geraintluff:
So if the host provides a buffer that's too small, should the plugin return the size we _wants_ the URI, not the actual size it's returning?

abique:
> So if the host provides a buffer that's too small, should the plugin return the size we _wants_ the URI, not the actual size it's returning?

Yes.
URI: 64 bytes
get_uri(capacity: 24) -> returns 64.

abique:
```C++
std::string get_dyn_str(std::function<int32_t (char *, uint32_t)> f)
{
   const auto sz = f(nullptr, 0);
   std::string s;
   s.resize(sz);
   f(s.data(), sz);
   return s;
}

auto uri = get_dyn_str([&] (char *s, uint32_t c) -> int32_t { return webview->get_uri(plugin, s, c); });
```

This is more or less what I have in mind on how to use this.
Once you wire it well, it is easy to work with.

geraintluff:
OK, got it!  A null `uri` is unexpected enough (for me) that I've add a note in the comments.

abique:
Oh sorry, I've forgotten to merge!

geraintluff:
This is awesome, thanks so much.

I don't know if you saw, but WCLAPs got a small shout-out in one of the ADC'25 talk proposals (about the future of audio plugins on the web).

PR #489, opened 2026-02-14
Introduce the octave-number extension

baconpaul:
Allowing the name of note 60 to be specified, per comments in extension header

### DIFF
--- a/include/clap/ext/draft/octave-number.h
+++ b/include/clap/ext/draft/octave-number.h
@@ -1,0 +1,30 @@
+#pragma once
+
+#include "../../plugin.h"
+
+// This extension allows a host to communicate to a plugin the
+// octave number associated with MIDI Note 60 (aka "Middle C").
+// Since various hosts and standards call note 60 either C3, C4 or C5,
+// host displays and plugin displays often mismatch absent this information
+// being shared.
+
+static CLAP_CONSTEXPR const char CLAP_EXT_PROJECT_LOCATION[] = "clap.octave-number/1";
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct clap_plugin_octave_number {
+   // If displaying the name of a note, and using standard CDEFGAB scale note names
+   // on a piano-roll-like display, this api will tell the plugin which octave number is
+   // associated with note 60.  For instance, calling this with '3' would indicate to
+   // the plugin that the consistent name for note 60 is "C3", for 72 is "C4" etc...
+   // if using CDEFGAB note names across a 12 note display.
+   // [main-thread]
+   void(CLAP_ABI *set_note60_octave)(const clap_plugin_t                  *plugin,
+                                     int octaveNumber);
+} clap_plugin_octave_number_t;
+
+#ifdef __cplusplus
+}
+#endif
--- a/include/clap/ext/draft/octave-number.h
+++ b/include/clap/ext/draft/octave-number.h
@@ -8,7 +8,7 @@
 // host displays and plugin displays often mismatch absent this information
 // being shared.
 
-static CLAP_CONSTEXPR const char CLAP_EXT_PROJECT_LOCATION[] = "clap.octave-number/1";
+static CLAP_CONSTEXPR const char CLAP_EXT_OCTAVE_NUMBER[] = "clap.octave-number/1";
 
 #ifdef __cplusplus
 extern "C" {
@@ -22,7 +22,7 @@ typedef struct clap_plugin_octave_number {
    // if using CDEFGAB note names across a 12 note display.
    // [main-thread]
    void(CLAP_ABI *set_note60_octave)(const clap_plugin_t                  *plugin,
-                                     int octaveNumber);
+                                     int8_t                                octaveNumber);
 } clap_plugin_octave_number_t;
 
 #ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


include/clap/ext/draft/octave-number.h
```diff
@@ -0,0 +1,30 @@
...
+
+typedef struct clap_plugin_octave_number {
+   // If displaying the name of a note, and using standard CDEFGAB scale note names
+   // on a piano-roll-like display, this api will tell the plugin which octave number is
+   // associated with note 60.  For instance, calling this with '3' would indicate to
+   // the plugin that the consistent name for note 60 is "C3", for 72 is "C4" etc...
+   // if using CDEFGAB note names across a 12 note display.
+   // [main-thread]
+   void(CLAP_ABI *set_note60_octave)(const clap_plugin_t                  *plugin,
+                                     int octaveNumber);
```
messmerd:
```suggestion
   void(CLAP_ABI *set_note60_octave)(const clap_plugin_t *plugin,
                                     int8_t               octave_number);
```
Using a fixed-width integer type + naming/style conventions

include/clap/ext/draft/octave-number.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "../../plugin.h"
+
+// This extension allows a host to communicate to a plugin the
+// octave number associated with MIDI Note 60 (aka "Middle C").
+// Since various hosts and standards call note 60 either C3, C4 or C5,
+// host displays and plugin displays often mismatch absent this information
+// being shared.
+
+static CLAP_CONSTEXPR const char CLAP_EXT_PROJECT_LOCATION[] = "clap.octave-number/1";
```
messmerd:
```suggestion
static CLAP_CONSTEXPR const char CLAP_EXT_OCTAVE_NUMBER[] = "clap.octave-number/1";
```

baconpaul: Oh that's embarrassing. Thanks!

madah81pnz1:
Very useful!

In addition, there could be a host API for the plugin to call the host. Since I assume not all plugins will use this, but the host has no other choice but to send this to all plugins otherwise.

Do we need a way for the host to reset this, e.g. a user configures something like 4 in their DAW and then decides to go back to "let the plugins decide".

Another function to choose preference for # or b, or if the note B should be shown as H, etc. But then it turns into another extension, a "locale" for musical symbols and notations.

baconpaul:
Right I figured the host would use implements-the-extension as an indication to send it to the plugin, but we could go theater direction. And that meant on host-side reset it could just call the plugin api again. I think the case where the daw has no opinion is not really a likely one. If it is ever called thats the daw opinion for the rest of time unless it is called again with a new opinion.

Yeah #/b/H etc... that's a totally different thing - basically a note-name-from-host. This is just solving the singular problem of what are my octave numbers.

include/clap/ext/draft/octave-number.h
messmerd: Looks like you're still using the camel-case `octaveNumber` instead of the snake case `octave_number` for the parameter name. And there's also a lot of extra spaces between the type and the parameter name, though that's not as important.

messmerd (approved):
Looks good to me. Just need to target the `next` branch rather than `main` before you merge.

abique:
LGTM.

Notes:
- The state can `Unknown->Known` but it can't `Known->Unknown`.
- the clap helper could optionally include some fallback code to determine what the host does:
```C++
std::optional<int> determineOctaveNumber(const clap_host *host)
{
   if (!strcmp(host->name, "xyz"))
      return 6;
   ...
   return {};
}

// plugin init code
...
if (!this->hostOctaveNumber.has_value())
   this->hostOctaveNumber = determineOctaveNumber(host);
```
- Please, can you include version.h and changelog update, so I can easily merge and release? :-)
- Please, can you provide the clap-helpers PR as well? :-)

baconpaul:
> LGTM.
> 
> Notes:
> 
> * The state can `Unknown->Known` but it can't `Known->Unknown`.

Yeah this is sort of on purpose. I don't know what it means for a DAW to have and then not have an opinion on this

> * the clap helper could optionally include some fallback code to determine what the host does:

We talked about this on discord some. Apparently most daws have an edit preference for this so turns out writing this code usefully is rather hard; and maintaining it is harder. (Like if bitwig adds an option in 6.1 or what not). I think we probably don't want that, although I had the same thought.

>  * Please, can you include version.h and changelog update, so I can easily merge and release? :-)
> * Please, can you provide the clap-helpers PR as well? :-)

Yup will do clap now and helpers later this morning.

abique:
> We talked about this on discord some. Apparently most daws have an edit preference for this so turns out writing this code usefully is rather hard; and maintaining it is harder. (Like if bitwig adds an option in 6.1 or what not). I think we probably don't want that, although I had the same thought.

For Bitwig it never changed as far as I know, and it could include a version range.
So it'd be an array of `{name, min_version, max_version}`.

baconpaul:
> > We talked about this on discord some. Apparently most daws have an edit preference for this so turns out writing this code usefully is rather hard; and maintaining it is harder. (Like if bitwig adds an option in 6.1 or what not). I think we probably don't want that, although I had the same thought.
> 
> For Bitwig it never changed as far as I know, and it could include a version range. So it'd be an array of `{name, min_version, max_version}`.

yeah bitwig is like that; i think logic is too. turns out cubase, s1, reaper have user options. as does MTS.

So just seemed too hairy

Pushed the version and changelog changes

PR #491, opened 2026-02-21
CLAP 1.2.8

abique:


mkruselj:
Mmm, I think Reaper allows any signed 32 bit int to be the octave offset here, haha. So using int_8t for the offset might not be enough (although practically it is). I'll check this and report back. Probably also makes sense for Reaper devs to clamp this option to more reasonable values (basically -1, 0 and 1)...

baconpaul:
> Mmm, I think Reaper allows any signed 32 bit int to be the octave offset here, haha. So using int_8t for the offset might not be enough (although practically it is). I'll check this and report back. Probably also makes sense for Reaper devs to clamp this option to more reasonable values (basically -1, 0 and 1)...

the value is the octave number so '3/4/5' for 'c3/c4/c5'

if someone wants to label note 60 as c130 or c-7842 then yeah uhhh this extension will wrap

geraintluff:
Perhaps could have a comment to say "it's recommended to set the octave number before querying the `clap.note-name` extension" in case this changes that one.

But LGTM either way.

mkruselj:
Looks like Reaper's octave offset range is -10000000...10000000 😁

defiantnerd (approved):
Looks good for me.

baconpaul:
I’ll add the comment and update helpers tomorrow 

I’ll add that it’s very likely a plugin implementing note name will change names and call changed when this is called if it implements note name. So daw writers probably want to call this before a note name scan and handle

baconpaul:
> Looks like Reaper's octave offset range is -10000000...10000000 😁

As long as A97583 is 440 and not 432, that’s fine with me

Bremmers (approved):

Trinitou:
As it's the first version in 2026, the "compat" extension and factory IDs should be removed, shouldn't they?

baconpaul:
> As it's the first version in 2026, the "compat" extension and factory IDs should be removed, shouldn't they?

I would sooner not do that on a minor point release

abique:
Thank you for the reminder!

I also prefer to delay it further, maybe this summer will be more chilled :-)

baconpaul:
ugh i ran out of time to add the comment and update helpers. Still on my list but need a day or two more still.

abique:
No rush :peace_symbol:

defiantnerd (approved):
Looks all fine for me.

baconpaul (approved):

PR #493, opened 2026-03-01
ext/params: clarify flush() dispatch expectations when active

Yashika0724:
This updates the documentation around `flush()` and `request_flush()` in `ext/params.h` to clarify the intended threading expectations when a plugin is active but the audio engine is temporarily idle (for example, when the transport is stopped).

The existing comments already state that `flush()` must not run concurrently with `process()`. This change makes that requirement more explicit in the “active but not processing” case to avoid ambiguity about which thread should dispatch `flush()`.

No functional changes are introduced. There are no ABI, struct layout, or behavioral modifications — this is purely a documentation clarification.

### DIFF
--- a/include/clap/ext/params.h
+++ b/include/clap/ext/params.h
@@ -296,6 +296,13 @@ typedef struct clap_plugin_params {
    // that the plugin may use the sample offset in process(), while this information would be
    // lost within flush().
    //
+   // When the plugin is active, flush() must be called from the audio-thread, and the host
+   // must ensure the audio thread is not concurrently in process() when doing so. This holds
+   // even when the audio engine is idle (e.g. transport stopped): the host must not substitute
+   // the main-thread for the audio-thread simply because process() is not currently running.
+   // If the host cannot guarantee an available audio-thread context, it must defer the flush
+   // until the next process() call, which will deliver the parameter changes instead.
+   //
    // [active ? audio-thread : main-thread]
    void(CLAP_ABI *flush)(const clap_plugin_t        *plugin,
                          const clap_input_events_t  *in,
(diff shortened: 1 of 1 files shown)


Yashika0724:
Hi @abique,
While going through ext/params.h I thought the threading expectations around `flush()` in the “active but transport stopped” case could benefit from a bit more explicit wording.
This PR only aims to clarify that scenario and preserve the existing non-concurrency requirement between `flush()` and `process()`.

If you think the current wording is already clear enough, I’m happy to adjust or drop the change. 
Thank you for taking a look.

defiantnerd:
Usually an audio engine is still running, although transport is stopped. The term would be "processing is stopped" which process is not being called anymore for some time.

Yashika0724:
Thanks for the clarification, that makes sense.

I’ll update the wording to refer to “processing is stopped” instead of “transport is stopped” to better match the intended terminology.

Yashika0724:
Thanks!
I’ve updated the wording to use “processing is stopped” instead.

include/clap/ext/params.h
```diff
@@ -370,8 +377,16 @@ typedef struct clap_host_params {
    // - clap_plugin.process()
    // - clap_plugin_params.flush()
    //
-   // This function is always safe to use and should not be called from an [audio-thread] as the
```
abique: For me the previous paragraph said everything and shortly.
We could replace the *should not* by *must not*.

abique:
@Yashika0724 thank you again :-)

@defiantnerd @baconpaul what do you think of this PR?

You know that I prefer shorter documentation where you have the strict minimum info and you can deduce everything else.

But overtime, we moved to a more detailed style of documentation and I'm probably not the right person to comment on that PR, so let me know what you think please :-)

baconpaul:
So the audio thread is a single thread so I don’t understand the comment that process can’t be concurrent with flush when active since they are both audio thread and there’s only one audio thread. So I don’t think the first clause is that useful

the second clause really says “active but not currently processing is still active and when active this is documented as audio thread”. I think that’s a fine clarification as a “for avoidance of doubt” type thing. I would say “for avoidance of doubt an active plugin must still use the audio thread for this even if not processing. This is only available as a main thread function for plugins which are not activated by the host”.

PR #494, opened 2026-03-01
adding state-progress extension

defiantnerd:
Introducing state-progress extension. This extension should improve the user experience when a plugin needs
time to restore the state, like loading sample files etc.

The idea is that the plugin calls `progress()`during state-extension's `load` call. When the thread is complete, the `done()` function should be called from the plugin in main-thread.

### DIFF
--- a/include/clap/ext/draft/state-progress.h
+++ b/include/clap/ext/draft/state-progress.h
@@ -1,0 +1,36 @@
+#pragma once
+
+#include "../state.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+  
+// This extension allows the plugin to use delayed state restoration and does not want to block
+// the main thread. The state restoration should be handled in a different thread so the user
+// experience is not impacted by the loading time. The plugin can then report its progress to the
+// host and notify it when it's done, so the host can update its UI accordingly.
+//
+// By this the host can disable the play button until all plugins are ready to play, and show a
+// progress bar if it wants to. This is especially useful for large sample-based plugins that need
+// to load a lot of data to restore their state, which can take a significant amount of time and
+// impact the user experience if done in the main thread.
+
+static CLAP_CONSTEXPR const char CLAP_EXT_EXTENSIBLE_AUDIO_PORTS[] = "clap.state-progress/draft-0";
+
+typedef struct clap_host_state_progress {
+   // informs the host that the plugin will restore its state in a different thread.
+   // the progress shall be updated in main thread using the progress callback.
+   // the host might want to display the progress that is expressed by "steps of total".
+   // If the plugin can not determine those steps, it should set steps and total to -1
+   // [main-thread]
+   void(CLAP_ABI *progress)(const clap_host_t *host, int32_t steps, int32_t total);
+
+   // informs the host that the plugin has restored it's state and is ready to play 
+   // [main-thread]
+   void(CLAP_ABI *done)(const clap_host_t *host);
+} clap_host_state_progress_t;
+
+#ifdef __cplusplus
+}
+#endif
--- a/include/clap/ext/draft/state-progress.h
+++ b/include/clap/ext/draft/state-progress.h
@@ -6,28 +6,33 @@
 extern "C" {
 #endif
   
-// This extension allows the plugin to use delayed state restoration and does not want to block
-// the main thread. The state restoration should be handled in a different thread so the user
-// experience is not impacted by the loading time. The plugin can then report its progress to the
+// This extension allows the plugin to use delayed state restoration while not blocking
+// the main thread. The state restoration must be handled in a different thread so the user
+// experience is not impacted by the loading time. The plugin shall report its progress to the
 // host and notify it when it's done, so the host can update its UI accordingly.
 //
-// By this the host can disable the play button until all plugins are ready to play, and show a
-// progress bar if it wants to. This is especially useful for large sample-based plugins that need
-// to load a lot of data to restore their state, which can take a significant amount of time and
-// impact the user experience if done in the main thread.
+// By this the host can disable the play button until all plugins are ready to play, and show one
+// or more progress bars if it wants to. This is especially useful for large sample-based plugins
+// that need to load a lot of data to restore their state, which can take a significant amount of
+// time and impact the user experience if done in the main thread.
 
-static CLAP_CONSTEXPR const char CLAP_EXT_EXTENSIBLE_AUDIO_PORTS[] = "clap.state-progress/draft-0";
+static CLAP_CONSTEXPR const char CLAP_EXT_STATE_PROGRESS[] = "clap.state-progress/draft-0";
 
 typedef struct clap_host_state_progress {
    // informs the host that the plugin will restore its state in a different thread.
    // the progress shall be updated in main thread using the progress callback.
    // the host might want to display the progress that is expressed by "steps of total".
    // If the plugin can not determine those steps, it should set steps and total to -1
-   // [main-thread]
+   // In this case the plugin might not update the host until the restore task is done.
+   // 
+   // Note that the first call must be during the state->load() call in the main thread.
+   // 
+   // [thread-safe]
    void(CLAP_ABI *progress)(const clap_host_t *host, int32_t steps, int32_t total);
 
-   // informs the host that the plugin has restored it's state and is ready to play 
-   // [main-thread]
+   // informs the host that the plugin has restored it's state and is ready to play.
+   // 
+   // [thread-safe]
    void(CLAP_ABI *done)(const clap_host_t *host);
 } clap_host_state_progress_t;
 
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -22,7 +22,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
         with:
           submodules: true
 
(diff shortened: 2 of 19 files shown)


include/clap/ext/draft/state-progress.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "../state.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+  
+// This extension allows the plugin to use delayed state restoration and does not want to block
+// the main thread. The state restoration should be handled in a different thread so the user
```
messmerd: The first sentence is not grammatically correct. "This extension allows [...] and does not want to [...]"

include/clap/ext/draft/state-progress.h
```diff
@@ -0,0 +1,36 @@
...
+
+static CLAP_CONSTEXPR const char CLAP_EXT_EXTENSIBLE_AUDIO_PORTS[] = "clap.state-progress/draft-0";
+
+typedef struct clap_host_state_progress {
+   // informs the host that the plugin will restore its state in a different thread.
+   // the progress shall be updated in main thread using the progress callback.
+   // the host might want to display the progress that is expressed by "steps of total".
+   // If the plugin can not determine those steps, it should set steps and total to -1
+   // [main-thread]
+   void(CLAP_ABI *progress)(const clap_host_t *host, int32_t steps, int32_t total);
```
messmerd, line 31: How would progress updates work if more than one plugin is restoring its state this way concurrently?

You'd probably need the host to generate a unique `clap_id` to associate with the task:
```c
// Informs the host that the plugin will restore its state in a different thread.
// If the plugin intends to send progress updates via the progress() function,
// it must set will_send_progress_updates to true.
// Returns a unique id to associate with the state load task, or CLAP_INVALID_ID
// if an error occurred.
// [main-thread & during-state-load]
clap_id(CLAP_API *init)(const clap_host_t *host, bool will_send_progress_updates);

```

Then the `progress()` function would look like this:
```c
void(CLAP_ABI *progress)(const clap_host_t *host, clap_id task, uint32_t steps, uint32_t total);
```
Note: It never sends `-1` for steps or total now, since that has been handled by `init()`, so I changed the type to the unsigned `uint32_t`.

And similarly, for the `done()` function:
```c
void(CLAP_ABI *done)(const clap_host_t *host, clap_id task);
```

include/clap/ext/draft/state-progress.h
```diff
@@ -0,0 +1,36 @@
...
+// the main thread. The state restoration should be handled in a different thread so the user
+// experience is not impacted by the loading time. The plugin can then report its progress to the
+// host and notify it when it's done, so the host can update its UI accordingly.
+//
+// By this the host can disable the play button until all plugins are ready to play, and show a
+// progress bar if it wants to. This is especially useful for large sample-based plugins that need
+// to load a lot of data to restore their state, which can take a significant amount of time and
+// impact the user experience if done in the main thread.
+
+static CLAP_CONSTEXPR const char CLAP_EXT_EXTENSIBLE_AUDIO_PORTS[] = "clap.state-progress/draft-0";
```
messmerd:
```suggestion
static CLAP_CONSTEXPR const char CLAP_EXT_STATE_PROGRESS[] = "clap.state-progress/draft-0";
```

include/clap/ext/draft/state-progress.h
messmerd, line 1: This extension could be turned into a generic way of making synchronous API functions asynchronous if there are any potential use cases for it beyond state loading.

If so, we may want to add `cancel` and `get_result` functions to support cancellation and retrieving the result of API functions that return a value.

include/clap/ext/draft/state-progress.h
messmerd: Also, how is the plugin expected to make main-thread progress updates? By calling `host->request_callback()` from its state loading thread, then calling `clap_host_state_progress.progress()` during `plugin.on_main_thread()`?

If `progress()` and `done()` were thread-safe, they could be called directly from the state loading thread. That would be easier and less error-prone for plugin developers since it shifts the extra complexity to the host side.

include/clap/ext/draft/state-progress.h
defiantnerd: Yesterday it seemed to make sense :D will rephrase it.

include/clap/ext/draft/state-progress.h
defiantnerd: Plugins that support delay loading already have a timer for their own UI. The idea was that it is easier for the host to maintain. Of course we also could shift this to the host. Or declaring it thread-safe, because the host will most likely have mechanisms for this anyway.

include/clap/ext/draft/state-progress.h
defiantnerd: thanks, copy and paste error.

include/clap/ext/draft/state-progress.h
defiantnerd: I had another extension in mind that allows the plugin to request a generic background task item. This can be combined, but allow this one to be focused on state restoration.

include/clap/ext/draft/state-progress.h
```diff
@@ -0,0 +1,41 @@
...
+// time and impact the user experience if done in the main thread.
+
+static CLAP_CONSTEXPR const char CLAP_EXT_STATE_PROGRESS[] = "clap.state-progress/draft-0";
+
+typedef struct clap_host_state_progress {
+   // informs the host that the plugin will restore its state in a different thread.
+   // the progress shall be updated in main thread using the progress callback.
+   // the host might want to display the progress that is expressed by "steps of total".
+   // If the plugin can not determine those steps, it should set steps and total to -1
+   // In this case the plugin might not update the host until the restore task is done.
```
Trinitou, line 26: - maybe the only condition should be `total == -1` -> `steps` is ignored
  - swap them in the signature: `progess(host,total,steps)`
- wouldn't it suffice to send the total number only once and then send the steps via a separate function
  ```
  start_progress(host, total)
  update_progress(host, steps)
  ```
  -> if `total == -1` -> `update_progress` must not be called

include/clap/ext/draft/state-progress.h
```diff
@@ -0,0 +1,41 @@
...
+   // 
+   // Note that the first call must be during the state->load() call in the main thread.
```
Trinitou, line 28: maybe we should think about and document the case the host calls `state->load()` again before the plugin calls `done()`
- I think this should be possible e.g. if
  - the user loads a host preset via the host (which contains the plugin state) and then during the long loading time decides to switch to another host preset
  - or maybe think of Bitwig's preset preview where a preset starts loading as soon as it gets selected in the browser ... so cycling through (expensive) host presets should not block the UI
- I think if all calls of this extension are forced to the main thread it would be possible to just allow the host to just call `state->load` at anytime and the previously running progress will be considered cancelled automatically

include/clap/ext/draft/state-progress.h
```diff
@@ -0,0 +1,41 @@
...
+   //
```
Trinitou, line 29: + I think a host maybe wants to offer a "Cancel" button in its UI to enable the user to stop long-running state loading

-> should be possible by adding a `clap_plugin_state_progress->cancel_progress` method?

Trinitou, line 29: my previous comment brings me to another point -> I think this extension should include a progress for the the `preset-load` extension as well as it's almost the same.
- by allowing the plugin to call `progress()` from within `clap_plugin_preset_load->from_location` as well?

Trinitou (changes requested):

include/clap/ext/draft/state-progress.h
Trinitou: by "host preset" I mean the host wrapping the plugin state into it's own custom plugin preset format (like Bitwig does + adds modulators + post fx chain)

include/clap/ext/draft/state-progress.h
```diff
@@ -6,28 +6,33 @@
...
+   // 
+   // [thread-safe]
    void(CLAP_ABI *progress)(const clap_host_t *host, int32_t steps, int32_t total);
 
-   // informs the host that the plugin has restored it's state and is ready to play 
-   // [main-thread]
+   // informs the host that the plugin has restored it's state and is ready to play.
+   // 
+   // [thread-safe]
    void(CLAP_ABI *done)(const clap_host_t *host);
```
Quant1um, line 36: What should happen if an error happens while loading? Maybe the extension should have a mechanism similar to `on_error` function in `preset-load` extension

include/clap/ext/draft/state-progress.h
```diff
@@ -6,28 +6,33 @@
...
 
 typedef struct clap_host_state_progress {
    // informs the host that the plugin will restore its state in a different thread.
    // the progress shall be updated in main thread using the progress callback.
    // the host might want to display the progress that is expressed by "steps of total".
    // If the plugin can not determine those steps, it should set steps and total to -1
-   // [main-thread]
+   // In this case the plugin might not update the host until the restore task is done.
+   // 
+   // Note that the first call must be during the state->load() call in the main thread.
```
Quant1um, line 28: I think it would be better if we had separate `start_load` function in `clap_plugin_state_progress` instead of reusing the existing `load` function to do state loading differently depending on the presence of a host extension

abique:
Thank you for the PR @defiantnerd ,

Here's a suggestion:

`progress()` could take a single value, if a `float`/`double` then between `0` to `1`, if `uint8_t` then between `0` and `UINT8_MAX`.
That way you prevent the total number of steps from varying between calls or even be invalid (<=0).

Saying that the progress report is between `0` and `100` would be fine I guess for most UI feedback, this is maybe the better call.

A progress call with the max value can be checked to determine the `done()` event. So `done()` isn't necessary maybe?

abique:
In regards to cancellation, `progress()` could return `true` or `false` to continue or cancel.
Another option would be a second method `bool should_continue()`?

PR #496, opened 2026-03-03
ext/preset-load: fix stale function name in on_error comment

Yashika0724:
In clap_host_preset_load, the on_error comment refers to clap_plugin_preset_load.load(), but the struct defines from_location().

This updates the comment to reference the correct function.

No functional or ABI changes.


Yashika0724:
Hi @abique,
While reviewing preset-load.h, I noticed the on_error comment still referenced load() instead of from_location().
This PR just updates that reference.

Thanks for taking a look.

baconpaul:
I definitely appreciate these careful reads and fixes to the comments!

abique:
Thank you again!

PR #497, opened 2026-03-06
ext/thread-pool: fix incorrect example code

Yashika0724:
While reading thread-pool.h I noticed the example code checks an `exec` field and calls `request_exec` with an extra argument.
The actual `clap_host_thread_pool` struct only defines `request_exec(const clap_host_t *host, uint32_t num_tasks)`.

This updates the example so it matches the current API.

No functional or ABI changes.

### DIFF
--- a/include/clap/ext/thread-pool.h
+++ b/include/clap/ext/thread-pool.h
@@ -22,8 +22,8 @@
 /// {
 ///    ...
 ///    bool didComputeVoices = false;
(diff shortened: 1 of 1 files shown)


Yashika0724:
Hi @abiague,
I noticed the example in thread-pool.h didn't quite match the current `clap_host_thread_pool` signature, so I updated it to reflect the actual API.
Thanks for taking a look.

abique:
Thank you very much, sorry for the delay!

PR #498, opened 2026-03-07
fix(surround): fix docs referencing non-existent struct members

Yashika0724:
include/clap/ext/surround.h describes a workflow using functions that don't exist in the actual structs:

clap_plugin_surround->changed() is mentioned, but the struct only has is_channel_mask_supported and get_channel_map.

clap_host_surround->get_preferred_channel_map() is mentioned, but the struct only has changed.

The documented host-initiated flow is currently impossible to implement. This PR updates the comment to reflect the actual API (deactivate -> get_channel_map -> activate).

No ABI/API changes.

### DIFF
--- a/include/clap/ext/surround.h
+++ b/include/clap/ext/surround.h
@@ -12,11 +12,8 @@
 //
 // If the host decides to change the project's surround setup:
 // 1. deactivate the plugin
-// 2. host calls clap_plugin_surround->changed()
-// 3. plugin calls clap_host_surround->get_preferred_channel_map()
-// 4. plugin eventually calls clap_host_surround->changed()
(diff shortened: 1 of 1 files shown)


Yashika0724:
Hi @abique ,

While reading through the surround extension I noticed that the workflow comment references a couple of functions that aren't present in the corresponding structs.

This change just updates the comment so the function names match the actual API.

No behavioral changes.
Happy to adjust if there’s a preferred wording.

abique:
Hi @Yashika0724 ,

Thank you very much for your PR.
Though, your new steps aren't correct.
The host should use `clap_plugin_configurable_audio_ports` to push the surround configuration.

I take care of this adjustment.

Issue #486, opened 2026-02-01
Clarify the meaning of some process statuses


Quant1um:
```
   // Processing succeeded, keep processing if the output is not quiet.
   CLAP_PROCESS_CONTINUE_IF_NOT_QUIET = 2,
```

How should a host determine when the plugin output is "quiet"? Should it use the output `constant_mask` information, check the output buffer for non-zero samples, or is it left unspecified?

```
   // Rely upon the plugin's tail to determine if the plugin should continue to process.
   // see clap_plugin_tail
   CLAP_PROCESS_TAIL = 3,
```

If latency is present, does tail time include latency or not? Is the tail of 0 and CLAP_PROCESS_TAIL equivalent to CLAP_PROCESS_SLEEP? (does the tail require the input to be quiet for the length of the tail for plugin to be allowed to sleep, or only the lack of "variation" is required)

I think these questions should be answered explicitly in the comments to avoid any possible misinterpretation of the standard

baconpaul:
My understanding, but I agree they should be explicit

1. A plugin which returns CLAP_PROCESS_TAIL for the number of samples returned by `tail` extension can be silenced. If the `tail` extension isn't implemented that is equivalent to infinite tail so CLAP_PROCESS_TAIL becomes CLAP_PROCESS_CONTINUE. If tail is 0 then PROCESS_TAIL and PROCESS_SLEEP are equivalent yes.

2. Quiet is left to the host implementer right now. I personally think something like 'entire output of process leaves the output buffer unchanged' is too much, but 'output buffer has an absmax of 1e-20' could be good. 

Curious if others have thoughts.

abique:
If you are implementing a plugin, ideally you only use continue or sleep.
Quiet let's the host determine (based on the host rules only) if the plugin's output is silent. This should be done by calculating the output volume and see if it was quiet enough for long enough. For example less than -96 dB for 1 second.
Tail, is useful when porting to clap. Essentially if the latency is M and the tail N samples, then if you feed some sounds to the audio effect, you'll need M+N output samples until the output becomes silent. The tail is also useful for bouncing.

Again, if possible return either continue or sleep, that way you're in total control.

PR #500, opened 2026-04-08
Clarify process status and tail extension docs

Quant1um:
Clarify process status related documentation in `process.h` and `ext/tail.h`, based on #486.

Unresolved questions:
- Should tail be latency-aware or not? This PR defines the tail as the amount of time in samples since last variation in the input, which implies latency-unaware-ness. This seems to be consistent with definitions given in #414. A different comment left in #486 seems to imply that you have to wait for M+N (where M is latency and N is tail) samples until the output is silent.

Corrections are appreciated!

### DIFF
--- a/include/clap/ext/tail.h
+++ b/include/clap/ext/tail.h
@@ -6,10 +6,9 @@
 ///
 /// This extension allows a plugin to specify its tail length,
 /// which is defined as the time in samples it takes for a plugin to stop producing non-silent
-/// output after receiving no non-silent input and no events (see CLAP_PROCESS_SLEEP for resume
-/// conditions).
+/// output after receiving only quiet input samples and no events.
 ///
-/// A host can use this information to determine when it is safe to stop processing without causing
+/// A host can use this information to determine when it can stop processing without causing
 /// a significant truncation of the audio output (see CLAP_PROCESS_TAIL), or to determine the total
 /// length of a track including reverb and delay tails.
 
--- a/include/clap/ext/tail.h
+++ b/include/clap/ext/tail.h
@@ -5,15 +5,13 @@
 /// @page tail
 ///
 /// This extension allows a plugin to specify its tail length,
-/// which is defined as the time it takes for a plugin to stop producing non-silent output after
-/// receiving no non-silent input and no events.
+/// which is defined as the time in samples it takes for a plugin to stop producing non-silent
+/// output after receiving no non-silent input and no events (see CLAP_PROCESS_SLEEP for resume
+/// conditions).
 ///
 /// A host can use this information to determine when it is safe to stop processing without causing
 /// a significant truncation of the audio output (see CLAP_PROCESS_TAIL), or to determine the total
 /// length of a track including reverb and delay tails.
-///
-/// Reported tail length is latency-unaware: if a plugin reports a non-zero latency, then the tail
-/// length must also include that latency (otherwise the host may truncate the output early).
 
 static CLAP_CONSTEXPR const char CLAP_EXT_TAIL[] = "clap.tail";
 
--- a/include/clap/ext/tail.h
+++ b/include/clap/ext/tail.h
@@ -2,6 +2,19 @@
 
 #include "../plugin.h"
 
+/// @page tail
+///
+/// This extension allows a plugin to specify its tail length,
+/// which is defined as the time it takes for a plugin to stop producing non-silent output after
+/// receiving no non-silent input and no events.
+///
+/// A host can use this information to determine when it is safe to stop processing without causing
+/// a significant truncation of the audio output (see CLAP_PROCESS_TAIL), or to determine the total
+/// length of a track including reverb and delay tails.
+///
+/// Reported tail length is latency-unaware: if a plugin reports a non-zero latency, then the tail
+/// length must also include that latency (otherwise the host may truncate the output early).
+
 static CLAP_CONSTEXPR const char CLAP_EXT_TAIL[] = "clap.tail";
 
 #ifdef __cplusplus
--- a/include/clap/process.h
+++ b/include/clap/process.h
@@ -20,7 +20,8 @@ enum {
    CLAP_PROCESS_CONTINUE_IF_NOT_QUIET = 2,
 
    // Rely upon the plugin's tail to determine if the plugin should continue to process:
-   // - it is safe to stop processing if the last variation (see CLAP_PROCESS_SLEEP) has occurred no later than clap_plugin_tail.get samples ago
+   // - it is safe to stop processing if the last input event/variation (see CLAP_PROCESS_SLEEP for resume conditions) 
+   //   has occurred at least clap_plugin_tail.get samples ago
    // - a tail of 0 is equivalent to CLAP_PROCESS_SLEEP
    // - an infinite tail is equivalent to CLAP_PROCESS_CONTINUE
    // 
--- a/include/clap/process.h
+++ b/include/clap/process.h
@@ -19,18 +19,15 @@ enum {
    // (a reasonable implementation could check if the output volume is less than a threshold for an extended period of time)
    CLAP_PROCESS_CONTINUE_IF_NOT_QUIET = 2,
 
-   // Rely upon the plugin's tail to determine if the plugin should continue to process:
-   // - it is safe to stop processing if the last input event/variation (see CLAP_PROCESS_SLEEP for resume conditions) 
-   //   has occurred at least clap_plugin_tail.get samples ago
-   // - it is safe to stop processing if no input events/variations have occurred 
-   //   since the last call to plugin->reset() or plugin->activate()
-   // - a tail of 0 is equivalent to CLAP_PROCESS_SLEEP
-   // - an infinite tail is equivalent to CLAP_PROCESS_CONTINUE
-   // 
-   // If either host or plugin do not implement the tail extension, this status is equivalent to CLAP_PROCESS_CONTINUE.
+   // Rely upon the plugin's tail to determine if the host should continue to call process.
+   //
+   // Stopping processing after at least [clap_plugin_tail.get] samples of quiet input and no events is expected to not cause significant truncation of the audio output.
+   // Calling plugin->reset() or plugin->activate() resets the tail and the plugin is expected to produce silence until the next event or variation in audio input.
+   //
+   // If the host does not support the tail extension, it is up to host to determine if it should continue processing or not.
    CLAP_PROCESS_TAIL = 3,
 
-   // Processing succeeded, it is safe to stop processing without causing a significant truncation of the audio output.
+   // Processing succeeded, stopping processing at this point is expected to not cause significant truncation of the audio output.
    //
    // Processing can be resumed by either:
    //  - a next event (note, parameter change, etc.)
--- a/include/clap/process.h
+++ b/include/clap/process.h
@@ -15,14 +15,24 @@ enum {
    CLAP_PROCESS_CONTINUE = 1,
 
    // Processing succeeded, keep processing if the output is not quiet.
+   // It is entirely up to host to determine what "quiet" means 
+   // (a reasonable implementation could check if the output volume is less than a threshold for an extended period of time)
    CLAP_PROCESS_CONTINUE_IF_NOT_QUIET = 2,
 
-   // Rely upon the plugin's tail to determine if the plugin should continue to process.
-   // see clap_plugin_tail
+   // Rely upon the plugin's tail to determine if the plugin should continue to process:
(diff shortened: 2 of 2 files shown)


defiantnerd:
Tail never contains the latency. Those are separete values. If they would be intertwened, communication with the host about it would be a mess.

giladkah:
## Review: Tail & Latency Documentation

This PR attempts to clarify the tail extension, but **there's a critical contradiction** that needs resolving first.

### 🚨 Blocking Issue: Contradicts Maintainer

The PR documentation states:
> "if a plugin reports a non-zero latency, then the tail length must also include that latency"

But the maintainer explicitly clarified in the comments:
> "Tail never contains the latency. Those are separate values."

**These are opposite directions.** The current text would instruct hosts and plugins to combine latency+tail, but the maintainer is saying they should be kept separate. This needs to be resolved before merging.

### What needs fixing

1. **Rewrite the latency clause** — clarify whether plugins or hosts are responsible for combining the values
2. **Define "variation"** — used throughout but never formally defined
3. **Fix temporal logic** — "no later than X ago" is backwards and confusing
4. **Add the unit** — tail.h doesn't state that values are in samples
5. **Edge cases** — DC offset thresholds, denormal handling, infinite-tail sentinel value

### Also noted

- 21 additional findings about unclear wording, edge cases, and incomplete documentation (see inline comments)
- No test vectors or reference implementation for the latency+tail contract
- Trailing whitespace issues

**Recommendation:** Resolve the latency contradiction with the maintainer first, then address the technical clarity issues.

include/clap/ext/tail.h
```diff
@@ -2,6 +2,19 @@
...
+/// @page tail
+///
+/// This extension allows a plugin to specify its tail length,
+/// which is defined as the time it takes for a plugin to stop producing non-silent output after
+/// receiving no non-silent input and no events.
+///
+/// A host can use this information to determine when it is safe to stop processing without causing
+/// a significant truncation of the audio output (see CLAP_PROCESS_TAIL), or to determine the total
+/// length of a track including reverb and delay tails.
+///
```
giladkah: 🟠 **Documentation contradicts maintainer intent: tail must NOT include latency**

The new text states 'the tail length must also include that latency (otherwise the host may truncate the output early).' This directly contradicts the maintainer's explicit statement: 'Tail never contains the latency. Those are separate values.' The phrase 'latency-unaware' is also self-contradictory here — if tail is latency-unaware, plugins should NOT bake latency into it. The intent appears to be: hosts must add latency on top of tail themselves, not that plugins add latency into the tail value. The sentence needs to be rewritten to say the host accounts for latency separately.

include/clap/process.h
```diff
@@ -15,14 +15,24 @@ enum {
    CLAP_PROCESS_CONTINUE = 1,
 
    // Processing succeeded, keep processing if the output is not quiet.
+   // It is entirely up to host to determine what "quiet" means 
+   // (a reasonable implementation could check if the output volume is less than a threshold for an extended period of time)
    CLAP_PROCESS_CONTINUE_IF_NOT_QUIET = 2,
 
-   // Rely upon the plugin's tail to determine if the plugin should continue to process.
-   // see clap_plugin_tail
+   // Rely upon the plugin's tail to determine if the plugin should continue to process:
```
giladkah: 🟠 **Logic inversion in CLAP_PROCESS_TAIL stop condition**

'safe to stop processing if the last variation has occurred no later than clap_plugin_tail.get samples ago' is logically backwards. 'No later than X samples ago' means the variation happened RECENTLY (within the last X samples). Processing should stop when enough time has elapsed, i.e., the last variation was AT LEAST X samples ago. Should read: 'at least clap_plugin_tail.get samples ago'.

include/clap/ext/tail.h
giladkah: 🟠 **Contradiction between new documentation and maintainer's stated design intent on latency**

PR context states the maintainer said 'Tail never contains the latency. Those are separate values.' The new doc says the opposite: 'if a plugin reports a non-zero latency, then the tail length must also include that latency.' These positions are irreconcilable. The API contract question — does tail include latency or not — is unresolved and must be settled before merge.

giladkah: 🟠 **Oxymoronic phrasing: 'latency-unaware' while simultaneously requiring latency be included**

'Reported tail length is latency-unaware' immediately followed by 'the tail length must also include that latency' is self-contradictory. If latency must be baked into the reported value, that value is latency-aware. The qualifier must be corrected to match whichever direction the contract actually specifies.

giladkah: 🟠 **Core latency claim directly contradicts the maintainer's stated design intent**

The new doc says 'if a plugin reports a non-zero latency, then the tail length must also include that latency' — but the maintainer explicitly stated 'Tail never contains the latency. Those are separate values.' This PR purports to clarify the latency question but resolves it in the direction opposite to the authoritative comment. A host reading this header will implement a combining strategy that contradicts the protocol design.

include/clap/ext/tail.h
```diff
@@ -2,6 +2,19 @@
...
 
+/// @page tail
+///
+/// This extension allows a plugin to specify its tail length,
+/// which is defined as the time it takes for a plugin to stop producing non-silent output after
+/// receiving no non-silent input and no events.
+///
+/// A host can use this information to determine when it is safe to stop processing without causing
+/// a significant truncation of the audio output (see CLAP_PROCESS_TAIL), or to determine the total
+/// length of a track including reverb and delay tails.
```
giladkah, line 13: 🟠 **'Latency-unaware' is inverted phrasing that will confuse implementers**

'Reported tail length is latency-unaware' reads as: the value is ignorant of latency (i.e., does NOT include it). But the very next sentence instructs plugins to ADD latency into the tail. These two halves of the sentence contradict each other. If the intent is 'the host won't compensate, so the plugin must', that needs to be stated directly — not buried in a negation that implies the opposite.

giladkah, line 13: 🟠 **Self-contradictory latency clause — contradicts maintainer and industry convention**

The sentence 'Reported tail length is latency-unaware: if a plugin reports a non-zero latency, then the tail length must also include that latency' is internally contradictory. 'Latency-unaware' means the tail value does not encode latency; the colon clause then says the plugin MUST bake latency into the tail value. These two halves negate each other. The maintainer explicitly stated 'Tail never contains the latency. Those are separate values.' The correct interpretation is that tail and latency are reported separately and the HOST is responsible for adding them. The clause starting with 'if a plugin reports a non-zero latency...' should either be removed or rewritten to describe host-side summation, e.g. 'Hosts that compensate for latency must add the plugin latency to the tail when computing the total safe-stop window.'

include/clap/ext/tail.h
```diff
@@ -2,6 +2,19 @@
 
 #include "../plugin.h"
 
+/// @page tail
```
giladkah, line 5: 🟠 **Tail unit (samples) not stated in tail.h**

process.h correctly specifies tail in samples ('clap_plugin_tail.get samples ago'), but the tail.h page-level documentation never states the unit. Audio engineers reading only the extension header will not know whether the return value is samples, frames, milliseconds, or seconds. The definition should explicitly say 'measured in samples at the current sample rate'.

defiantnerd:
Can you please stop this slop?

Quant1um:
> Tail never contains the latency. Those are separete values. If they would be intertwened, communication with the host about it would be a mess.

Got it! That is my understanding as well, that tail is just the number of samples from an event/non-zero audio sample to the point where it is safe to stop processing, and I wrote the docs from the point of view of that assumption. The thing I got confused by is the comment left in #486

edwloef:
I think it would be good to add a short note specifying whether the host should "forget" about previous inputs when the plugin's `reset` function is called, wrt. deciding whether to send a plugin to sleep in response to `CLAP_PROCESS_TAIL`. Maybe something like `it is safe to stop processing if the last input event/variation (see CLAP_PROCESS_SLEEP for resume conditions) has occurred at least clap_plugin_tail.get samples ago, or the last input event/variation occured before the last call to plugin->reset`?

include/clap/ext/tail.h
```diff
@@ -2,6 +2,17 @@
...
+///
+/// This extension allows a plugin to specify its tail length,
+/// which is defined as the time in samples it takes for a plugin to stop producing non-silent
+/// output after receiving no non-silent input and no events (see CLAP_PROCESS_SLEEP for resume
```
abique: You can replace "no non-silent" by "quiet".

abique: Maybe "(see CLAP_PROCESS_SLEEP for resume conditions)" isn't needed here.

include/clap/ext/tail.h
```diff
@@ -2,6 +2,17 @@
...
+/// conditions).
+///
+/// A host can use this information to determine when it is safe to stop processing without causing
```
abique: Incorrect use of the word "safe".

include/clap/process.h
```diff
@@ -15,14 +15,27 @@ enum {
...
+   // - it is safe to stop processing if the last input event/variation (see CLAP_PROCESS_SLEEP for resume conditions)
```
abique: Incorrect use of "safe" again.

I don't think the list helps.

If the host doesn't support tail, then it is up to the host to determine what to do. It would reasonable to fallback to `CONTINUE_IF_NOT_QUIET`.

include/clap/process.h
```diff
@@ -15,14 +15,27 @@ enum {
...
+   //   since the last call to plugin->reset() or plugin->activate()
+   // - a tail of 0 is equivalent to CLAP_PROCESS_SLEEP
+   // - an infinite tail is equivalent to CLAP_PROCESS_CONTINUE
+   // 
+   // If either host or plugin do not implement the tail extension, this status is equivalent to CLAP_PROCESS_CONTINUE.
    CLAP_PROCESS_TAIL = 3,
 
-   // Processing succeeded, but no more processing is required,
-   // until the next event or variation in audio input.
+   // Processing succeeded, it is safe to stop processing without causing a significant truncation of the audio output.
```
abique: Incorrect use of "safe".

abique (changes requested):

edwloef:
Oh, while we're here, could we also get some clarification on what `main-thread,audio-thread` means for `clap_plugin_tail->get`? Should that mean or be changed to `thread-safe` or `active ? audio-thread : main-thread`? In general, what does the return value represent when the plugin is deactivated, or should that also be a disallowed scenario, like with the `latency` extension?

Quant1um:
> Oh, while we're here, could we also get some clarification on what `main-thread,audio-thread` means for `clap_plugin_tail->get`? Should that mean or be changed to `thread-safe` or `active ? audio-thread : main-thread`? In general, what does the return value represent when the plugin is deactivated, or should that also be a disallowed scenario, like with the `latency` extension?

I agree that `[main-thread,audio-thread]` is a bit weird and it should probably be changed to `[thread-safe]` or, even better, `[audio-thread]` or even `[active & audio-thread]`, I am not sure what the right choice would be and changing it could result in breakage with existing plugins/hosts. More discussion is needed probably.

edwloef:
Does/should a plugin calling `clap_host.request_process` count as a change in input for the tail extension? I.e. when a plugin calls `clap_host.request_process` and returns `CLAP_PROCESS_TAIL`, should the host keep calling `process` for `clap_plugin_tail.get` samples? The added documentation for `CLAP_PROCESS_SLEEP` explicitly mentions `clap_host.request_process`, but the added documentation for `CLAP_PROCESS_TAIL` does not.

Quant1um:
> Does/should a plugin calling `clap_host.request_process` count as a change in input for the tail extension? I.e. when a plugin calls `clap_host.request_process` and returns `CLAP_PROCESS_TAIL`, should the host keep calling `process` for `clap_plugin_tail.get` samples? The added documentation for `CLAP_PROCESS_SLEEP` explicitly mentions `clap_host.request_process`, but the added documentation for `CLAP_PROCESS_TAIL` does not.

I'd say it should count as a resume condition for tail as well (for consistency with SLEEP), I agree that the docs should be clearer.

Quant1um:
Also I think `tail->get` should be changed to `[thread-safe & active]`, it should be a relatively safe change as it already doesn't make sense to call `tail->get` before sample rate is available. Thoughts?

edwloef:
Perhaps `CLAP_PROCESS_CONTINUE_IF_NOT_QUIET` should also explicitly mention whether it's latency-aware. i.e. whether when the host should keep processing for at least `latency` more samples to determine whether "quiet" has been achieved, or whether the plugin should keep track of that.

PR #505, opened 2026-05-14
Definition for GUI with UIKIT (iOS/iPadOS)

defiantnerd:
This adds a identification string:

`static const CLAP_CONSTEXPR char CLAP_WINDOW_API_UIKIT[] = "uikit";`

and a new member `clap_uiview uikit;` which has no further meaning except to make the semantics clear.

This is needed for the clap-wrapper for auv3-ios.


abique:
This looks good to me, thanks!

PR #510, opened 2026-07-03
Clarify bypass parameter count and process() behavior

Schroedingers-Cat:
This PR clarifies that there can only be zero or one bypass parameter per plugin and that the value of the bypass parameter should not influence the host calling plugin->process() or not.


abique:
Thanks!

PR #512, opened 2026-07-17
Documentation: Add more information about how event addresses work wrt parameters

reuk:
The interaction between addressed and wildcard parameter/modulation events feels underspecified at the moment, so I've tried to make it clearer how plugins should respond to different sequences of events.

I'm not claiming to have got this correct, but hopefully this serves as a good starting point to critique/review and nudge in the right direction.

### DIFF
--- a/include/clap/events.h
+++ b/include/clap/events.h
@@ -105,6 +105,9 @@ enum {
    // In case of a concurrent global value/modulation versus a polyphonic one,
    // the voice should only use the polyphonic one and the polyphonic modulation
    // amount will already include the monophonic signal.
+   //
+   // See the section titled "Voice addressing" below for more information about
+   // the behaviour of parameter values and modulations.
    CLAP_EVENT_PARAM_VALUE = 5,
    CLAP_EVENT_PARAM_MOD = 6,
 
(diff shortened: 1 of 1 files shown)


reuk:
A question: How do event PCKN addresses interact with with `CLAP_PARAM_IS_AUTOMATABLE_PER_NOTE_ID`, `CLAP_PARAM_IS_MODULATABLE_PER_KEY`, and other similar flags from params.h? Is the plugin supposed to ignore an event with a non-wildcard value for a given address element if that element's flag is unset? Or is the host only allowed to send wildcard values for any address element with an unset flag? Or something else?

baconpaul:
They are really wildcards. The param methods are telling the host what it can expect a param to respond to. But if you have a by note id param and you send note id -1 it targets all playing voices

reuk:
OK thanks, it sounds like the plugin should just ignore an event with a non-wildcard note ID in the case that a given CLAP_PARAM_IS_AUTOMATABLE_PER_NOTE_ID is unset, then.

baconpaul:
Yeah I would say a host that sends that has a bug

